PR #3, opened 2016-07-16
Navigation drawer with client list added

vjs3:
This PR contains:
- Navigation Drawer with Homescreen.
- Client list with corresponding client id's for the respective self-service user.

### DIFF
--- a/app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
@@ -1,0 +1,138 @@
+package org.mifos.selfserviceapp.ui.activities;
+
+
+import android.os.Bundle;
+import android.support.design.widget.NavigationView;
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentTransaction;
+import android.support.v4.view.GravityCompat;
+import android.support.v4.widget.DrawerLayout;
+import android.support.v7.app.ActionBarDrawerToggle;
+
+import android.view.MenuItem;
+import android.view.View;
+import android.support.v7.widget.Toolbar;
+
+import org.mifos.selfserviceapp.R;
+import org.mifos.selfserviceapp.ui.fragments.ClientListFragment;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+
+/**
+ * @author Vishwajeet
+ * @since 14/07/2016
+ */
+
+public class HomeActivity extends BaseActivity implements NavigationView.OnNavigationItemSelectedListener {
+
+    private NavigationView mNavigationView;
+    private DrawerLayout mDrawerLayout;
+
+    @BindView(R.id.toolbar)
+    Toolbar toolbar;
+
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
```diff
@@ -0,0 +1,138 @@
...
+        setContentView(R.layout.activity_home);
+
+        ButterKnife.bind(this);
+
+        if (toolbar != null) {
+            setSupportActionBar(toolbar);
+            toolbar.setTitle(getString(R.string.home));
+        }
+
+        replaceFragment(new ClientListFragment(), false, R.id.container);
```
ishan1604: Don't create fragments like this. 
Use getInstance() method to create Fragments.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/ClientListFragment.java
```diff
@@ -0,0 +1,145 @@
...
+        this.clientList = clientList;
+        inflateClientList();
+        if (swipeRefreshLayout.isRefreshing())
+            swipeRefreshLayout.setRefreshing(false);
+    }
+
+    /**
+     * setting adapter to display list of clients on the client list screen
+     */
+    private void inflateClientList() {
```
ishan1604: don't initialise the adapter again and again on refresh. 
if the adapter is already there, just call notifyDataSetChanged(), it'll take care of refreshing your list.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/ClientListFragment.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void showErrorFetchingClients(String message) {
+        //TODO: Handle this error properly
+        Toast.makeText(getActivity(), message, Toast.LENGTH_LONG).show();
+    }
+
+    @Override
+    public void showInternalServerError(String message) {
```
ishan1604: You are doing the same task in two different methods.

ishan1604:
Please fix these problems and resend PR.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/ClientListFragment.java
vjs3: @ishan1604 I am showing different error message in both the methods so that user can know the reason of failure in loading client list. One method is showing error of its caused due to internal server issue, other if there is any authentication error or some error related to user end, depending upon the response code from server.

ishan1604: @vjs3 That is fine. I get that you are showing different error messages which is very nice and makes the user experience much better.

But I don't understand the need of two different methods for this. 

Can't we just have one method on the view called ```showError(String message)```

and then from within the presenter you can call it whenever you have an error
like ```getMvpView().showError("Failed to load client list")``` or ```getMvpView().showError("Could not authenticate user")``` etc.

vjs3: Oh. Yes didn't think that way. Thanks, will do that.

PR #7, opened 2016-08-12
Recent transactions feature added

vjs3:
This PR contains:

- Recent transaction features in which client can see list of their recent transactions.
- It will consist of date of transaction, amount transacted and type of transaction value i.e, if it is Waive charge or pay charge etc.
- Added icons for navigation drawer. All the icons were taken from google material design site, hence completely open source. Link to icons: https://design.google.com/icons/


app/src/main/java/org/mifos/selfserviceapp/api/services/RecentTransactionsService.java
```diff
@@ -0,0 +1,17 @@
...
+import retrofit2.http.GET;
+import retrofit2.http.Path;
+
+/**
+ * @author Vishwajeet
+ * @since 10/08/2016
+ */
+public interface RecentTransactionsService {
+    @GET(ApiEndPoints.CLIENTS + "/{clientId}/transactions")
+    Call<Transaction> getRecentTransactionsList(@Path("clientId") int clientId);
```
satyan: this can be moved to the clients service itself..

app/src/main/java/org/mifos/selfserviceapp/api/DataManager.java
```diff
@@ -49,6 +50,10 @@ public DataManager(PrefManager prefManager) {
         return baseApiManager.getLoanAccountsListApi().getLoanAccountsList(id);
     }
 
+    public Call<Transaction> getRecentTransactions(int id) {
```
satyan: rename id to clientId

app/src/main/java/org/mifos/selfserviceapp/data/Transaction.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author Vishwajeet
+ * @since 10/8/16.
+ */
+public class Transaction {
+    private int id;
+    private List<Transaction> pageItems = new ArrayList<Transaction>();
```
satyan: same refactoring needed as accounts.. move this list to something like TransactionListResponse class.

app/src/main/java/org/mifos/selfserviceapp/data/Transaction.java
```diff
@@ -0,0 +1,70 @@
...
+    private List<Integer> submittedOnDate = new ArrayList<Integer>();
```
satyan: why is date stored as integer? If it's timestamps, it's generally "long" type.

app/src/main/java/org/mifos/selfserviceapp/data/Transaction.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public Type getType() {
+        return type;
+    }
+
+    public void setType (Type type) {
+        this.type = type;
+    }
+
+    public class Type {
```
satyan: can make this a static class.  The advantage is that you don't need an instance of the outer class to access this.

app/src/main/java/org/mifos/selfserviceapp/presenters/RecentTransactionsPresenter.java
```diff
@@ -0,0 +1,75 @@
...
+        Call<Transaction> call = dataManager.getRecentTransactions(clientId);
+        getMvpView().showProgress();
+
+        call.enqueue(new Callback<Transaction>() {
+            @Override
+            public void onResponse(Response<Transaction> response) {
+                getMvpView().hideProgress();
+
+                if (response.code() == 200) {
+                    Transaction recentTransaction = response.body();
```
satyan: Is this variable being used?

app/src/main/java/org/mifos/selfserviceapp/data/Transaction.java
vjs3: yes!I was waiting for your review of accounts refactoring, will do it now.

PR #12, opened 2016-08-18
Saving-account detail feature added

vjs3:
This PR contains:

- By tapping on any of the list item of saving accounts list, client can now see details of each of their saving accounts.

- With this feature, clients can see details of  such as product name, account balance, total deposits, nominal interest rate, min. required balance etc for each of their saving accounts.

### DIFF
--- a/app/src/main/java/org/mifos/selfserviceapp/presenters/SavingAccountsDetailPresenter.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/presenters/SavingAccountsDetailPresenter.java
@@ -1,0 +1,75 @@
+package org.mifos.selfserviceapp.presenters;
+
+import android.content.Context;
+
+import org.mifos.selfserviceapp.R;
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mifos/selfserviceapp/presenters/SavingAccountsDetailPresenter.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public void loadSavingAccountDetails(int accountId) {
+        Call<SavingAccount> call = dataManager.getSavingAccountDetails(accountId);
+        getMvpView().showProgress();
+        call.enqueue(new Callback<SavingAccount>() {
+            @Override
+            public void onResponse(Response<SavingAccount> response) {
+                getMvpView().hideProgress();
+
+                if (response.code() == 200) {
```
satyan: think about a way to generalize this logic. seems to be duplicated everywhere..

PR #17, opened 2016-08-23
Fund tranfer feature with in own accounts has been added

vjs3:
With this PR user can now transfer their funds to their own saving or loan accounts.

### DIFF
--- a/app/src/main/java/org/mifos/selfserviceapp/api/ApiEndPoints.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/api/ApiEndPoints.java
@@ -12,4 +12,5 @@ public class ApiEndPoints {
     public static final String AUTHENTICATION = "authentication";
     public static final String CLIENTS = "clients";
     public static final String SAVINGSACOUNTS = "savingsaccounts";
+    public static final String ACCOUNTTRANSFERS = "accounttransfers";
 }
--- a/app/src/main/java/org/mifos/selfserviceapp/data/FundTransfer/FundTransferTemplateResponse.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/data/FundTransfer/FundTransferTemplateResponse.java
@@ -1,0 +1,30 @@
+package org.mifos.selfserviceapp.data.FundTransfer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author Vishwajeet
+ * @since 23/8/16.
+ */
+
+public class FundTransferTemplateResponse {
+    private List<FundTransferTemplate> fromAccountOptions = new ArrayList<>();
+    private List<FundTransferTemplate> toAccountOptions = new ArrayList<>();
+
+    public List<FundTransferTemplate> getFromAccountOptions() {
+        return fromAccountOptions;
+    }
+
+    public void setFromAccountOptions(List<FundTransferTemplate> fromAccountOptions) {
+        this.fromAccountOptions = fromAccountOptions;
+    }
+
+    public List<FundTransferTemplate> getToAccountOptions() {
+        return toAccountOptions;
+    }
+
+    public void setToAccountOptions(List<FundTransferTemplate> toAccountOptions) {
+        this.toAccountOptions = toAccountOptions;
+    }
+}
--- a/app/src/main/java/org/mifos/selfserviceapp/presenters/FundTransferPresenter.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/presenters/FundTransferPresenter.java
@@ -1,0 +1,103 @@
+package org.mifos.selfserviceapp.presenters;
+
+import android.content.Context;
+
+import org.mifos.selfserviceapp.R;
+import org.mifos.selfserviceapp.api.DataManager;
+import org.mifos.selfserviceapp.data.FundTransfer.FundTransferRequest;
+import org.mifos.selfserviceapp.data.FundTransfer.FundTransferResponse;
+import org.mifos.selfserviceapp.data.FundTransfer.FundTransferTemplate;
+import org.mifos.selfserviceapp.data.FundTransfer.FundTransferTemplateResponse;
+import org.mifos.selfserviceapp.injection.ActivityContext;
+import org.mifos.selfserviceapp.presenters.base.BasePresenter;
+import org.mifos.selfserviceapp.ui.views.FundTransferView;
+
+import javax.inject.Inject;
+
+import retrofit2.Call;
+import retrofit2.Callback;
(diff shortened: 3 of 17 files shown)


app/src/main/java/org/mifos/selfserviceapp/api/ApiEndPoints.java
```diff
@@ -12,4 +12,5 @@
     public static final String AUTHENTICATION = "authentication";
     public static final String CLIENTS = "clients";
     public static final String SAVINGSACOUNTS = "savingsaccounts";
+    public static final String ACCOUNTTRANSFERS = "accounttransfers";
```
satyan: usually the style here is ACCOUNT_TRANSFERS which makes the name easier to read. without the _ it's a little difficult to understand.

app/src/main/java/org/mifos/selfserviceapp/data/FundTransfer/FundTransferTemplateResponse.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class FundTransferTemplateResponse {
+    private List<FundTransferTemplate> fromAccountOptions = new ArrayList<>();
+    private List<FundTransferTemplate> toAccountOptions = new ArrayList<>();
+
+    public List<FundTransferTemplate> getFromAccountOptions() {
+        return fromAccountOptions;
+    }
+
+    public void setFromAccountOptions(List<FundTransferTemplate> fromAccountOptions) {
```
satyan: nit: you may not need the set methods for these. Since these are usually populated by retrofit, you can just have the get methods here.

app/src/main/java/org/mifos/selfserviceapp/presenters/FundTransferPresenter.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public void submitTransfer(FundTransferRequest fundTransferRequest) {
+        Call<FundTransferResponse> call = dataManager.submitTransfer(fundTransferRequest);
+        getMvpView().showProgress();
+
+        call.enqueue(new Callback<FundTransferResponse>() {
+            @Override
+            public void onResponse(Response<FundTransferResponse> response) {
+                if (response.code() == 200) {
```
satyan: this logic is present everywhere. Figure out a way to reuse something.. Can be a separate PR with that change.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/FundTransferFragment.java
```diff
@@ -0,0 +1,209 @@
...
+        for (int i = 0; i < fromAccountNumber.size(); i++) {
+            fromAccountItems[i] = fromAccountNumber.get(i).getAccountNo();
+        }
+        ArrayAdapter<String> fromAccountAdapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_list_item_1, fromAccountItems);
+        spFromAccount.setAdapter(fromAccountAdapter);
+        spFromAccount.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+            @Override
+            public void onItemSelected(AdapterView<?> parent, View view, int position, long
+                    id) {
+                fromClientId = fundTransferTemplate.getFromAccountOptions().
```
satyan: need to validate the lists for data. Also use the fromAccountNumber list that you have above.

app/src/main/java/org/mifos/selfserviceapp/api/ApiEndPoints.java
vjs3: Completely agree. Mistake at my part. Will change it.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/FundTransferFragment.java
```diff
@@ -0,0 +1,209 @@
...
+        mFundTransferPresenter.attachView(this);
+
+        mFundTransferPresenter.loadAccountTransferTemplate();
+
+        return rootView;
+    }
+
+    @Override
+    public void showFundTransferTemplate(final FundTransferTemplateResponse fundTransferTemplate) {
+        List<FundTransferTemplate> fromAccountNumber = fundTransferTemplate.getFromAccountOptions();
```
satyan: It's useful to call the variable fromAccountNumberList to make it clear.. otherwise it looks like it's an account number.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/FundTransferFragment.java
```diff
@@ -0,0 +1,209 @@
...
+        for (int i = 0; i < toAccountNumber.size(); i++) {
+            toAccountItems[i] = toAccountNumber.get(i).getAccountNo();
+        }
+        ArrayAdapter<String> toAccountAdapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_list_item_1, toAccountItems);
+        spToAccount.setAdapter(toAccountAdapter);
+        spToAccount.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+            @Override
+            public void onItemSelected(AdapterView<?> parent, View view, int position, long
+                    id) {
+                toClientId = fundTransferTemplate.getToAccountOptions().
```
satyan: same as above. Validate data else it might be a candidate for runtime exceptions.

app/src/main/java/org/mifos/selfserviceapp/data/FundTransfer/FundTransferTemplateResponse.java
vjs3: Yeah. But I thought it goes with the way I am writing code. Does it make
any difference if yes then please tell me.

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/FundTransferFragment.java
vjs3: Yes. Validations are missing as I told you earlier. Will add them.

app/src/main/java/org/mifos/selfserviceapp/data/FundTransfer/FundTransferTemplateResponse.java
satyan: yeah. it's not something with this PR. In general, you expose methods that are intended to be used. In this case, set method is not intended to be used.

This is something you can keep in mind in general. Everything you expose a public method, think about whether it's needed..

You may choose to fix it in this PR, or tackle them generally in future PRs. Up to you.

PR #30, opened 2016-10-13
Fixes

puneetkohli521:


ishan1604:
The PR is not in a good shape.

- The PR title doesn't tell what it tries to improve.
- There are multiple fixes in this PR that are not related to each other and thus should be tackled in separate PRs
- Also, make sure when you send a PR, the commits are squashed otherwise it is very difficult to review the changes.
- Before committing the code, make sure you reformat it in the IDE.

PR #31, opened 2016-10-18
Refactored Model Directory name and PreferenceHelper name

therajanmaurya:


ishan1604:
Objects is not a very nice name for the package.

Issue #45, opened 2016-10-29
Add : LoanAccount Status (Active or not) in LoanAccount list item.


therajanmaurya:
In current user is able to see the LoanAccount list but there is no sign to see. what is the status of LoanAccount. it is Active or not.

How to achieve this issue : 
- Add a circular Image view in LoanAccount Item list in Adapter .
- show the status of the LoanAccount.

ashu-dadhich:
can I work on it @therajanmaurya

therajanmaurya:
Yeah,  Go ahead. In which issue you have to add image view in  LoanAccount Adapter.
and UI would be like this. I am talking about the Image for showing the status.
![group 13](https://cloud.githubusercontent.com/assets/5724482/19828665/32f5698a-9dea-11e6-83e4-296d99d73762.png)

ashu-dadhich:
Yeah got it @therajanmaurya  thanks

therajanmaurya:
@top-gun007 You were asking me, how will you get Account status. 
Please look into the LoanAccountListAdapter, you will found `private Status` field.
that will give you the status of the Account. 
Sorry, I have missed getter setter when I was fixing the POJO. So make the getter setter of the LoanAccountListAdapter and you are able to access the Status of the Account.

I want from this issue only : 
- Add Circular ImageView or you can use VIew Class instead because we want to fill the color in the view. So Please Use View.
  For Example : http://stackoverflow.com/a/14209836/3176277.
- Set any random color in Adapter that is already present in colors.xml. Do not need to add any exatra color.

**I will open separate issue for this setting correct color order for account in Adapter for both Savings and Loan.**

therajanmaurya:
Fixed in #54

PR #52, opened 2016-10-31
Show loan status with reference to issue #45 

ashu-dadhich:
Changes-

- added 8 xmls in the drawables for image status.

- added view in row_loan_account.xml

- checked multiple status conditions in LoanAccountListAdapter and show accordingly

- implemented getter and setter methods for status

### DIFF
--- a/app/src/main/java/org/mifos/selfserviceapp/models/accounts/loan/LoanAccount.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/models/accounts/loan/LoanAccount.java
@@ -50,6 +50,14 @@ public class LoanAccount implements Parcelable {
     @SerializedName("status")
     private Status status;
 
+    public Status getStatus() {
+        return status;
+    }
+
+    public void setStatus(Status status) {
+        this.status = status;
+    }
+
     @SerializedName("loanType")
     private LoanType loanType;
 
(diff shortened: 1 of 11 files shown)


ashu-dadhich:
@therajanmaurya please have a look

app/src/main/java/org/mifos/selfserviceapp/models/accounts/loan/LoanAccount.java
```diff
@@ -50,6 +50,14 @@
...
     private Status status;
 
+    public Status getStatus() {
+        return status;
+    }
+
+    public void setStatus(Status status) {
+        this.status = status;
+    }
+
```
therajanmaurya, line 104: Please add getter setter for others field also that does not have getter setter.

app/src/main/res/layout/row_loan_account.xml
```diff
@@ -1,14 +1,19 @@
...
-    android:paddingRight="@dimen/row_padding_right"
->
+    android:paddingRight="@dimen/row_padding_right">
+
+    <View
+        android:id="@+id/vw_loanStatus"
+        android:layout_width="25dp"
+        android:layout_height="25dp"
+        android:layout_gravity="center"
+        android:background="@drawable/closed" />
```
therajanmaurya, line 16: This is good but I found, as we will add account type. we have to add more shape drawable.
that is not good.

Can you, please change this View to circular ImageView using this https://github.com/openMF/android-client/blob/master/mifosng-android/src/main/java/com/mifos/mifosxdroid/views/CircularImageView.java
custom Circular ImageView.

and Add  CircularImageView in xml.

PR #71, opened 2016-12-10
RTL Support Addition

superav:
Layout files have been edited and changed so that they allow support for lanugauge scripts that read from right to left and those that read left to right.

### DIFF
--- a/app/src/main/res/layout/activity_loan_account_details.xml
+++ b/app/src/main/res/layout/activity_loan_account_details.xml
@@ -14,7 +14,7 @@
         android:layout_weight="1"
         android:gravity="center_vertical"
         android:orientation="horizontal"
-        android:paddingLeft="@dimen/row_padding_left"
+        android:paddingStart="@dimen/row_padding_left"
         android:paddingTop="@dimen/table_row_padding_top">
 
         <TextView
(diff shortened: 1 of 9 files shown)


app/src/main/res/layout/activity_loan_account_details.xml
```diff
@@ -15,6 +15,7 @@
         android:gravity="center_vertical"
         android:orientation="horizontal"
         android:paddingLeft="@dimen/row_padding_left"
```
tarun0: This line is redundant as the minApiVersion is 17. `android:paddingStart="@dimen/row_padding_left"` in itself is sufficient.

tarun0 (changes requested):
Missed [this](https://github.com/superav/self-service-app/blob/dbfc32e81774e89e5065f5b59400478dd64cfdb6/app/src/main/res/layout/activity_saving_account_details.xml#L35) maybe?

Please make these small changes and submit the task again.

tarun0:
@superav Please squash the 3 commits into a single commit. Then @therajanmaurya will review it finally before merging.

mdew192837 (commented):
@tarun0 I have a few questions. Just trying to learn. :) Should we leave paddingLeft for backwards compatibility? You mentioned miniApiVersion 17. Does that mean paddingLeft is not needed? Also, should we add android:supportsRtl="true" in the AndroidManifest.xml file?

tarun0:
@mdew192837 Yes, to support RTL, one must enable it in the manifest file by using `android:supportsRtl="true"`. Since the app's minSdkVersion is 17, it means that this app can't be installed in the Android versions below Jellybean (any API < 17). The attributes such as paddingLeft are used with API<17 (or if RTL is disabled in manifest). 
Since RTL is already enabled (this is why you can't find the change in the commits) and the minSdkVersion for the app is 17, we don't need marginLeft because marginLeft and marginStart will mean the same.

Hope that clarifies your query.

PR #72, opened 2016-12-11
Addition of RTL Support

superav:
These changes allow support of language scripts that read both right to left and left to right.


tarun0 (approved):
The PR seems fine to me. @therajanmaurya will approve it finally before merging.

Issue #64, opened 2016-11-25
Add feature to show user details.


therajanmaurya:
How to achieve this issue 
1. Add layout in Navigation Header to show logged in user profile image and name. 
2. Onclick profile ImageView opens a Layout that shows the user profile.
    like : User profile image, Name Account Number etc.

akmittal006:
how to get user image?

therajanmaurya:
Here is the API END point https://demo.openmf.org/api-docs/apiLive.htm#selfclientsimages. 
It is giving the Client Image Binary data.

akmittal006:
can i add picasso ??

therajanmaurya:
Yes you can. Go ahead.

joydeepsaha05:
@therajanmaurya Should the profile image and name be added on top of the current Mifos image?

hardik124:
I want to work on this.

m-murad:
Can I work on this

therajanmaurya:
@hardik124 Go ahead 👍. Please try to be quick.

therajanmaurya:
@hardik124  Any update

akmittal006:
I was already working on it . I had my mid terms, so i couldnt update. @therajanmaurya  any other issue i can work on?

![screenshot_4](https://cloud.githubusercontent.com/assets/8525715/23821652/77e1a4c2-065f-11e7-8e06-046348b16f1b.png)

hardik124:
@therajanmaurya  i am not able to retrieve user image please check #187

dilpreet2028:
Added in #338

PR #172, opened 2017-03-11
       feat:Add feature to show user details

hardik124:
1)Add layout in Navigation Header to show logged in user profile image and name.

              2)  Onclick profile ImageView opens a Layout that shows the user profile.

#64  #159 


app/src/main/java/org/mifos/selfserviceapp/api/services/ClientImageService.java
```diff
@@ -0,0 +1,17 @@
...
+import okhttp3.ResponseBody;
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Path;
+
+/**
+ * Created by f390 on 7/3/17.
+ */
+
+public interface ClientImageService {
```
therajanmaurya, line 14: Do not need to make new service for client image. Revert nack this and add this method in `ClientService`.
and also remove the code from `BaseApiManager` that you have added.

app/src/main/java/org/mifos/selfserviceapp/presenters/LoginPresenter.java
```diff
@@ -145,9 +145,9 @@ public void onError(Throwable e) {
                     public void onNext(Page<Client> clientPage) {
                         getMvpView().hideProgress();
                         if (clientPage.getPageItems().size() != 0) {
-                            long clientId = clientPage.getPageItems().get(0).getId();
-                            getMvpView().showClient(clientId);
-                            preferencesHelper.setClientId(clientId);
+                            Client clientCurrent = clientPage.getPageItems().get(0);
+                            getMvpView().showClient(clientCurrent);
```
therajanmaurya, line 149: I do not think. we need this changes because for making request we need only client id that is enough for us. do not need to pass whole client.

Please revert back.

app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
```diff
@@ -119,4 +182,77 @@ protected void onDestroy() {
...
+    private void openUserDetails(Bitmap image) {
+        Intent profilePage = new Intent(HomeActivity.this, UserDetails.class);
+        profilePage.putExtra(Constants.CURRENT_CLIENT, current_client);
+        profilePage.putExtra(Constants.CLIENT_IMAGE, image);
+        startActivity(profilePage);
+    }
+
+    void setUserImage() {
+        {
+            final Call<ResponseBody> imageResponse = dataManager.getClientImage(clientId);
```
therajanmaurya, line 212: You just broke the MVP architecure concept in HomeActivity. You have to make A HomeMvpView and HomePresenter and need to make logic separately.   

Please look other feature implementation

therajanmaurya (changes requested):

therajanmaurya:
@hardik124 You are not using the MVP architecture guidelines in Home and UserDetails Activity.
It's the MVP architecture that we are following strictly. Please look the other feature implementation.

app/src/main/java/org/mifos/selfserviceapp/presenters/LoginPresenter.java
hardik124: My approach was to get the client only once, otherwise if i pass only id, the HomeActivity will need to send a GET request again which will increase data usage.

app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
hardik124: Okay , i will implement that

app/src/main/java/org/mifos/selfserviceapp/api/services/ClientImageService.java
hardik124: Okay , i will implement this.

app/src/main/java/org/mifos/selfserviceapp/presenters/LoginPresenter.java
therajanmaurya: No, It's not. We are building the end user banking app. We need fresh data at every step.

hardik124: okay i will implement this.

hardik124: Should i pass client image to userDetails activity? or get it new again.
Also I think , user image size should be reduced to <100 kb , currently its 1MB+

therajanmaurya: You do not need to pass the client id any where. check the DataManager class. As you are logging into the app. client id is passing into the DataManager So make you retro call in DataManager using the DataManager's client id.

hardik124: Should I store clients image or retrieve it again for userDetails activity?

PR #290, opened 2017-04-07
feature: Implement user details

naman14:
Implement user details- Adds user name and image in navigation header

screenshot_20170407-133203-01

Also includes small fix for removing transactions menu item from transactions fragment - Transactions menu item is also visible in the transactions fragment itself. Remove the menu item when in savings account transactions fragment.

### DIFF
--- a/app/src/main/java/org/mifos/selfserviceapp/api/services/ClientService.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/api/services/ClientService.java
@@ -5,6 +5,7 @@ import org.mifos.selfserviceapp.models.Page;
 import org.mifos.selfserviceapp.models.client.Client;
 import org.mifos.selfserviceapp.models.client.ClientAccounts;
 
+import okhttp3.ResponseBody;
 import retrofit2.http.GET;
 import retrofit2.http.Path;
 import retrofit2.http.Query;
@@ -16,15 +17,23 @@ import rx.Observable;
  */
 public interface ClientService {
 
+    String CLIENT_ID = "clientId";
+
     //This is a default call and Loads client from 0 to 200
     @GET(ApiEndPoints.CLIENTS)
     Observable<Page<Client>> getClients();
 
-    @GET(ApiEndPoints.CLIENTS + "/{clientId}/accounts")
-    Observable<ClientAccounts> getClientAccounts(@Path("clientId") long clientId);
+    @GET(ApiEndPoints.CLIENTS + "/{clientId}")
+    Observable<Client> getClientForId(@Path(CLIENT_ID) long clientId);
+
+    @GET(ApiEndPoints.CLIENTS + "/{clientId}/images")
+    Observable<ResponseBody> getClientImage(@Path(CLIENT_ID) long clientId);
 
     @GET(ApiEndPoints.CLIENTS + "/{clientId}/accounts")
-    Observable<ClientAccounts> getAccounts(@Path("clientId") long clientId,
+    Observable<ClientAccounts> getClientAccounts(@Path(CLIENT_ID) long clientId);
+
+    @GET(ApiEndPoints.CLIENTS + "/{clientId}/accounts")
+    Observable<ClientAccounts> getAccounts(@Path(CLIENT_ID) long clientId,
             @Query("fields") String accountType);
 
 }
--- a/app/src/main/java/org/mifos/selfserviceapp/presenters/UserDetailsPresenter.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/presenters/UserDetailsPresenter.java
@@ -1,0 +1,125 @@
+package org.mifos.selfserviceapp.presenters;
+
+import android.content.Context;
+import android.graphics.Bitmap;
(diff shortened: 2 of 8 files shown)


app/src/main/java/org/mifos/selfserviceapp/api/services/ClientService.java
```diff
@@ -16,15 +17,23 @@
  */
 public interface ClientService {
 
+    String CLIENT_ID = "clientId";
```
droidchef, line 20: I am assuming you abstracted this out as a constant, so this should be ```private static final```. What do you think ?

app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
```diff
@@ -42,6 +49,12 @@
     @Inject
     PreferencesHelper preferencesHelper;
 
+    @Inject
+    UserDetailsPresenter detailsPresenter;
+
+    TextView username;
```
droidchef: I'd suggest suffixing class identifier in case of views, so it becomes more readable in later part of the code.

Something like ```tvUsername``` and ```ivUserProfilePicture``` etc.

And also the views can be made ```private``` unless you have a plan to access them from outside this class.

What do you think?

app/src/main/java/org/mifos/selfserviceapp/presenters/UserDetailsPresenter.java
```diff
@@ -0,0 +1,127 @@
...
+
+                            //removing 'data:image/jpg;base64' from the response
+                            //the response is of the form of 'data:image/jpg;base64, .....'
+                            final String pureBase64Encoded =
+                                    encodedString.substring(encodedString.indexOf(',') + 1);
+
+                            final byte[] decodedBytes =
+                                    Base64.decode(pureBase64Encoded, Base64.DEFAULT);
+
+                            Bitmap decodedBitmap =
```
droidchef, line 113: I am not very familiar with RxJava concepts. So I have this doubt about, on which thread will this code be called?

I see that you are observing on ```AndroidSchedulers.mainThread()```, I don't think it is a good idea to decode the bitmap on the main thread.

Please throw some light on this, thanks!

droidchef (changes requested):

app/src/main/java/org/mifos/selfserviceapp/api/services/ClientService.java
naman14: Checkstyle test was failing since string "clientId" was being repeated more than 4 times here, so I made it as a constant.
Interface constants are implicitly `public static final` so defining them as `static final` is redundant. Also, interfaces cannot have private modifiers.

app/src/main/java/org/mifos/selfserviceapp/ui/activities/HomeActivity.java
naman14: Good point, will update the naming of views and yeah the views can be made private

app/src/main/java/org/mifos/selfserviceapp/presenters/UserDetailsPresenter.java
naman14: Yeah, looks like this will run on the main thread, looking into the optimal implementation for this

app/src/main/java/org/mifos/selfserviceapp/api/services/ClientService.java
droidchef: Ahh My bad, I didn't realize it was an interface. 

Old habits of writing Retrofit Services as inner classes :( 

Sorry about this!

naman14:
Added a new commit with the requested changes

app/src/main/java/org/mifos/selfserviceapp/presenters/UserDetailsPresenter.java
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                })
+        );
+    }
+
+    public void getUserImage() {
+        checkViewAttached();
+        subscriptions.add(dataManager.getClientImage()
+                .observeOn(Schedulers.newThread())
+                .subscribeOn(Schedulers.io())
```
therajanmaurya, line 87: @naman14 observeOn will be on `AndroidSchedulers.mainThread()` and about the async loading image you can schedule that on newThread by subscribing to new thread `Schedulers.newThread()`

Note checkout this: https://github.com/ReactiveX/RxJava/wiki/The-RxJava-Android-Module#observing-on-the-ui-thread

therajanmaurya (changes requested):

therajanmaurya:
@naman14 make the above changes and Please remove the ic_icon.png with vector image.
Here is the tutorial: http://stackoverflow.com/a/35402459/3176277 and finally squash you commits and force push again.

**Note: can you please update you message and please mention which issue you are fixing like Fixes #{Issue number}. for future please follow PR template Thanks**

therajanmaurya (approved):

PR #304, opened 2017-06-03
feat: Update and Withdraw loan

### LINKED ISSUE #303: Loan updation implementation
#### Steps 
- To update a loan,a new fragment will be created which can use the exisiting fragment_add_application layout as the api allows to update all the elements provided in the UI part of this layout but with a new presenter.
- LoanAccount(model) needs to be passed during the intialization
- ProductId retrieved from loanAccount will be used to call the template which requires ProductId. API used for fetching template:- 
  https://demo.openmf.org/fineract-provider/api/v1/self/loans/template?templateType=individual&clientId=1&productId=1&tenantIdentifier=default&pretty=true
- Existing values from the retrieved LoanAccount(model) will be displayed after fetching loan template.
- After the self service user has made the required changes then existing LoanPayload model can be used for making a (PUT) request to:- https://demo.openmf.org/fineract-provider/api/v1/self/loans/{loanId} 

@therajanmaurya do I need to make any modifications in this implementation?

dilpreet2028:
Fixes #303 
#### `Loan Withdraw:`
                   

@therajanmaurya have a look at the PR


therajanmaurya:
@dilpreet96 Please fix the conflicts.

dilpreet96:
@therajanmaurya looking into it

dilpreet96:
@therajanmaurya conflicts fixed

dilpreet2028:
@therajanmaurya conflicts fixed

app/src/main/java/org/mifos/selfserviceapp/presenters/LoanAccountWithdrawPresenter.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public void attachView(LoanAccountWithdrawView mvpView) {
+        super.attachView(mvpView);
+    }
+
+    @Override
+    public void detachView() {
+        super.detachView();
+        subscriptions.unsubscribe();
```
therajanmaurya: use ``subscriptions.clear();``

app/src/main/java/org/mifos/selfserviceapp/presenters/LoanApplicationPresenter.java
```diff
@@ -63,13 +64,17 @@ public void onError(Throwable e) {
                     @Override
                     public void onNext(LoanTemplate loanTemplate) {
                         getMvpView().hideProgress();
-                        getMvpView().showLoanTemplate(loanTemplate);
+                        if (loanState == LoanState.CREATE) {
```
therajanmaurya, line 68: @dilpreet96 I think, Template would be same in both cases? So why two separate methods

app/src/main/java/org/mifos/selfserviceapp/presenters/LoanApplicationPresenter.java
```diff
@@ -89,7 +94,11 @@ public void onError(Throwable e) {
                     @Override
                     public void onNext(LoanTemplate loanTemplate) {
                         getMvpView().hideProgress();
-                        getMvpView().showLoanTemplateByProduct(loanTemplate);
+                        if (loanState == LoanState.CREATE) {
```
therajanmaurya, line 98: same as above

app/src/main/java/org/mifos/selfserviceapp/ui/enums/LoanState.java
```diff
@@ -0,0 +1,10 @@
+package org.mifos.selfserviceapp.ui.enums;
+
+/**
+ * Created by dilpreet on 3/6/17.
+ */
+
+public enum LoanState {
+    CREATE,
+    EDIT
```
therajanmaurya: @dilpreet96 you are using update everywhere So use here to ``UPDATE`` instead ``EDIT``

app/src/main/res/layout/fragment_loan_withdraw.xml
```diff
@@ -0,0 +1,69 @@
...
+            tools:text="@string/account_number"
+            android:textColor="@color/black"
+            android:textSize="@dimen/text_large"/>
+    </LinearLayout>
+
+    <EditText
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:id="@+id/et_withdraw_reason"
+        android:layout_marginTop="36dp"
```
therajanmaurya: @dilpreet96 you can do one thing here, define this margin in dimens.xml then you can use this dimens value other places, So the margin value becomes unique in the whole app.

app/src/main/java/org/mifos/selfserviceapp/presenters/LoanApplicationPresenter.java
dilpreet2028: @therajanmaurya In case of an update loan state after the loan template is loaded, current values are need to be set in the fields so I divided them into two functions one for new application where we only need to intialize template and second where we need to place the current values of the loan which needs to be updated after the template has been loaded.

app/src/main/java/org/mifos/selfserviceapp/presenters/LoanApplicationPresenter.java
dilpreet2028: @therajanmaurya In case of Loan Update,
We need to load the current value of loan purpose if it exists for the loan which is to be updated.
So `showUpdateLoanTemplateByProduct` load the spinner with all values from template and then check if it is being intialized for the first time and does it contains a loan purpose if then it yes loads those values otherwise it loads the values fetched from the template i.e. the principal amount ,currency and rest.

therajanmaurya:
make the changes and let me know.

dilpreet2028:
@therajanmaurya changes made

PR #328, opened 2017-06-29
refractor: Home Screen

dilpreet2028:
<b>Note : As the current  total savings balance is more than 12 digit value which would not be possible I think so modified savings total amount for demo purpose:  </b>

Demo with actual values (of savings total amount): 


dilpreet2028:
@therajanmaurya @puneetkohli521 please review this

ippezshelby:
Hi All, Can you please make changes to the login so that we can connect to a local instance of Fineract not online demo only for testing as sometimes there is not internet connectivity. Or this is possible??

dilpreet2028:
Hey @Ippezrobert , self service currently does not have an offline support i.e. it does not allows to connect to local instance of FIneract.

PR #338, opened 2017-07-10
feat: User details

dilpreet2028:
ss


app/src/main/java/org/mifos/selfserviceapp/ui/fragments/HomeFragment.java
```diff
@@ -127,6 +129,13 @@ public void showUserImage(final Bitmap bitmap) {
             public void run() {
                 ((ImageView) toolbarView.findViewById(R.id.iv_user_image)).
                         setImageBitmap(bitmap);
+                toolbarView.findViewById(R.id.iv_user_image).
+                        setOnClickListener(new View.OnClickListener() {
```
therajanmaurya, line 133: Use Butterknife @onClick annotation instead `setOnClickListener(..)`

therajanmaurya (changes requested):

app/src/main/java/org/mifos/selfserviceapp/ui/fragments/HomeFragment.java
dilpreet2028: @therajanmaurya as custom view in toolbar is added in onResume and removed at onPause state https://github.com/dilpreet96/self-service-app/blob/abb8736564ef83d009489ce19c05234953c7da88/app/src/main/java/org/mifos/selfserviceapp/ui/fragments/HomeFragment.java#L207
so, butterknife is not be able to find the image as it is added at runtime due to which I used the manual method

therajanmaurya (approved):

PR #402, opened 2017-09-06
fix: Invisible button text in MFDatePicker

### LINKED ISSUE #401: Buttons not displayed in calendar
**Summary:**
The OK and CANCEL buttons are not visible on the calendar as the text is white
**Steps to reproduce:**
Go to Add a new load and change submission date
**Expected behavior:**
The button text should be visible
**Observed behavior:**
Button text was invisible
Samsung Galaxy A5(2016)
**Screenshots:**
image

mayank8318:
Fixes #401 

Please Add Screenshots If any UI changes.
The buttons are now visible -
image

- [:+1: ] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [:+1: ] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [:+1: ] If you have multiple commits please combine them into one commit by squashing them.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -39,4 +39,7 @@
         <item name="android:padding">@dimen/layout_padding_8dp</item>
     </style>
 
+    <style name="MaterialDatePickerTheme" parent="Theme.AppCompat.Light.Dialog.Alert">
+        <item name="colorAccent">@color/accent</item>
+    </style>
 </resources>
--- a/app/src/main/java/org/mifos/selfserviceapp/utils/MFDatePicker.java
+++ b/app/src/main/java/org/mifos/selfserviceapp/utils/MFDatePicker.java
@@ -13,6 +13,8 @@ import android.support.v4.app.Fragment;
 import android.text.format.DateFormat;
 import android.widget.DatePicker;
 
+import org.mifos.selfserviceapp.R;
+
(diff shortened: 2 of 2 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -39,4 +39,7 @@
         <item name="android:padding">@dimen/layout_padding_8dp</item>
     </style>
 
+    <style name="DialogTheme" parent="Theme.AppCompat.Light.Dialog.Alert">
```
therajanmaurya: @mayank8318 Why don't you use this style https://github.com/mayank8318/self-service-app/blob/bd0c29186e5af18720ccfc8dcf7c246760c12f76/app/src/main/res/values/styles.xml#L28

therajanmaurya (changes requested):

app/src/main/res/values/styles.xml
mayank8318: @therajanmaurya that is using **colorAccent = primaryColor** but the dialog should use the **accent color**. I didn't change that because that style might be used somewhere else so I created a new style.
Should I make changes in that style itself?

therajanmaurya: @mayank8318 that style is using in Material Dialog builder class in /utils. So do one thing, rename the style with `MaterialDatePickerTheme` and we will enhance this style if needed.

PR #501, opened 2017-12-07
fix: Issue #447 - Display Currency Code if Currency Symbol is NULL

### LINKED ISSUE #447: Currency symbol displayed as null
**Summary:**

Currency symbol is displayed as null in Charges screen.

**Steps to reproduce:**

Open the charges tab.

**Expected behavior:**

It should have the appropriate currency sign.

**Observed behavior:**

null is displayed instead of the currency sign.

Samsung A5(2016)

**Screenshots:**

image

luckyman20:
Fixes #447 
@therajanmaurya @santoshmath Please review it!!

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/models/Currency.java
+++ b/app/src/main/java/org/mifos/mobilebanking/models/Currency.java
@@ -40,6 +40,9 @@ public class Currency implements Parcelable {
     }
 
     public String getDisplaySymbol() {
+        if (displaySymbol == null) {
+            return code;
+        }
         return displaySymbol;
     }
 


app/src/main/java/org/mifos/mobilebanking/models/Currency.java
```diff
@@ -40,6 +40,9 @@ public void setDecimalPlaces(int decimalPlaces) {
     }
 
     public String getDisplaySymbol() {
+        if (displaySymbol == null) {
+            displaySymbol = getCode();
+        }
```
therajanmaurya: @luckyman20 PR looks good, I want a small change 
```
 public String getDisplaySymbol() {
        if(displaySymbol == null) {
            return code;
        }
        return displaySymbol;
    }
```

therajanmaurya (changes requested):

app/src/main/java/org/mifos/mobilebanking/models/Currency.java
luckyman20: @therajanmaurya Made the changes

therajanmaurya (approved):

PR #584, opened 2018-02-07
Fix Issue#559 : Disabled the continue button in absence of internet

ivary43:
Fixes #{559}

- \[x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- \[x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- \[x] If you have multiple commits please combine them into one commit by squashing them.


ivary43:
@therajanmaurya, @dilpreet96  please review :)

dilpreet2028:
@ivary43 what if my internet connection stops working while I was entering the amount or remark?

ivary43:
@dilpreet96  if you lose internet connection at that point you will be able to go to the next page but won't be able to transfer.

ivary43:
@dilpreet96  any changes you want?

therajanmaurya:
@ivary43 This is not a good way to handle No Internet connection until then the steps need to fetch the data from the server each time. So here It is happening only once So checking at the final place makes more sense.

PR #646, opened 2018-02-17
Changed the error message if username or password did'nt matched

### LINKED ISSUE #645: Error message should be more helpful
On logging in , if user types incorrect password or email then the message is displayed about the api. It should be more helpful like incorrect email or password.

@dilpreet96 @therajanmaurya I'm working on this.

Rish02Sharma:
Fixes #645 

@dilpreet96 @therajanmaurya Review please. :)

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
+++ b/app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
@@ -195,9 +195,12 @@ public class LoginPresenter extends BasePresenter<LoginView> {
                     , resources.getString(R.string.password), resources.getInteger(R.integer.
                             password_minimum_length)));
             return false;
+        } else {
+            getMvpView().showMessage(context
+                    .getString(R.string.error_unauthorised));
+            return false;
         }
 
-        return true;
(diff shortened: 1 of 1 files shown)


ivary43 (commented):
@Rish02Sharma  please change your approach because the error message you are getting  is from the server side it's not generated from the app side plus you are editing the wrong function "isCredentialsValid" looks for credentials entered by the user is in correct format  or not it doesn't authenticate the credentials.

ivary43:
@Rish02Sharma you can't change the message because it is generated from the server side.

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
```diff
@@ -195,9 +195,12 @@ private boolean isCredentialsValid(final String username, final String password)
                     , resources.getString(R.string.password), resources.getInteger(R.integer.
                             password_minimum_length)));
             return false;
+        } else {
```
miPlodder, line 198: @Rish02Sharma Its wrong fix, if we do this we will always enter `else` and never call to server will be made.

PR #717, opened 2018-03-11
fix:added error message in case server is down

### LINKED ISSUE #588: No error message present while logging in if server are down.
**Summary:**

There should be error message showing the error why the user is not able to log in as per the status code received from the server side.

**Expected behaviour:**

The app should the error why the user is not able to log in (leaving wrong credentials) if it is due to the server side.

**Observed behaviour:**

No Such error message was present.

ivary43:
Fixes #588

- \[x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- \[x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- \[x] If you have multiple commits please combine them into one commit by squashing them.


ivary43:
@therajanmaurya , @dilpreet96  please review.

PR #759, opened 2018-03-18
fix: Keyboard persistence bug fixed

### LINKED ISSUE #752: Keyboard persists during 3rd party transfer
**Summary:**
Under 3rd party transfer when the remarks are entered and the review is selected , the keyboard doesn't get dismissed. 

**Steps to reproduce:**
Go to 3rd party transfer and add the remarks and select REVIEW.

**Screenshot**
device-2018-03-18-144442

**Expected behaviour:**
The keyboard should be dismissed.

sridharjajoo:
Fixes #752 

device-2018-03-18-231838

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/activities/base/BaseActivity.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/activities/base/BaseActivity.java
@@ -1,5 +1,6 @@
 package org.mifos.mobilebanking.ui.activities.base;
 
+import android.app.Activity;
 import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.Intent;
@@ -12,6 +13,7 @@ import android.support.v4.view.ViewCompat;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.Toolbar;
 import android.view.MenuItem;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.Toast;
 
 import org.mifos.mobilebanking.MifosSelfServiceApp;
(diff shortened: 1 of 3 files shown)


sridharjajoo:
@therajanmaurya @dilpreet96 Kindly review!

app/src/main/java/org/mifos/mobilebanking/utils/Utils.java
```diff
@@ -83,4 +85,12 @@ public static String formatTransactionType(String type) {
         }
         return builder.toString();
     }
+
+    public static void hideKeyboard(Activity activity) {
```
therajanmaurya: @sridharjajoo Please create this method in `BaseActivity` and use from there.

therajanmaurya (changes requested):

app/src/main/java/org/mifos/mobilebanking/ui/activities/base/BaseActivity.java
```diff
@@ -252,4 +254,12 @@ public int stackCount() {
         return getSupportFragmentManager().getBackStackEntryCount();
     }
 
+
+    public static void hideKeyboard(Activity activity) {
+        InputMethodManager inputMethodManager =
+                (InputMethodManager) activity.getSystemService(
```
therajanmaurya: Please use Context instead of Activity

therajanmaurya (changes requested):

app/src/main/java/org/mifos/mobilebanking/ui/activities/base/BaseActivity.java
sridharjajoo: Actually Activity is used for using the getCurrentFocus( ) . What I can do is pass context as the parameter and convert it to activity something like this:

 public static void hideKeyboard(Context context) {
      
  Activity activity = (Activity) context;

        InputMethodManager inputMethodManager =
                (InputMethodManager) context.getSystemService(
                        Activity.INPUT_METHOD_SERVICE);

        inputMethodManager.hideSoftInputFromWindow(
                activity.getCurrentFocus().getWindowToken(), 0);
    }

therajanmaurya (approved):

PR #760, opened 2018-03-18
fix: classifying accounts according to their types

dilpreet2028:

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/presenters/SavingsMakeTransferPresenter.java
+++ b/app/src/main/java/org/mifos/mobilebanking/presenters/SavingsMakeTransferPresenter.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import org.mifos.mobilebanking.R;
 import org.mifos.mobilebanking.api.DataManager;
 import org.mifos.mobilebanking.injection.ApplicationContext;
+import org.mifos.mobilebanking.models.payload.AccountDetail;
 import org.mifos.mobilebanking.models.templates.account.AccountOption;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mifos/mobilebanking/presenters/SavingsMakeTransferPresenter.java
```diff
@@ -97,20 +98,20 @@ public void onNext(AccountOptionsTemplate accountOptionsTemplate) {
...
+                .flatMap(new Function<AccountOption, Observable<AccountDetail>>() {
                     @Override
-                    public Observable<String> apply(AccountOption accountOption) {
-                        return Observable.just(accountOption.getAccountNo());
+                    public Observable<AccountDetail> apply(AccountOption accountOption) {
+                        return Observable.just(new AccountDetail(accountOption.getAccountNo(),
+                                accountOption.getAccountType().getValue()));
                     }
                 })
-                .distinct()
```
dilpreet2028, line 109: @therajanmaurya , this was the cause of the crash while trying to open transfers from a loan account as it was filtering and removing the accounts with the same account number, but as you can see in the image we have account number as 0000001 for saving account and also for loan account. 
As we are searching directly from the accountOption template here:https://github.com/openMF/mifos-mobile/pull/760/files#diff-fd4b24c3db287adc9a965db1dec74947R307 but we are setting the index obtained from this to the spinner which is attached to the filtered list due to which it was throwing indexoutofbound exception.

So, inorder to keep both the accounts number (same account but different type so I had to make a custom spinner adapter to differentiate between them

PR #765, opened 2018-03-19
Fix #747 : Added Checks for empty Edittexts and for validity of Url in Update Endpoints Setting

### LINKED ISSUE #747: No check for empty Edittexts and for validity of Url, causes App to crash
**Summary:**

No check for Empty Base Url, cause App to crashes

**Steps to reproduce:**

1) Go to Setting
2) Change the Base Url from Update Endpoints

**Expected behaviour:**

Instead User should be informed to enter valid Url

**Observed behaviour:**

App crashes and next time when the user tries to open it App crashes ( It degrades UX )

Xiaomi, Marshmallow

**Screenshots:**

issue_empty_endpoint

miPlodder:
Fixes #747 

pr_747

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -412,6 +412,8 @@
     <string name="language">Language</string>
     <string name="choose_language">Choose your language</string>
     <string name="notification">Notifications</string>
+    <string name="base_url">Base URL</string>
+    <string name="tenant">Tenant</string>
 
     <string-array name="faq_qs">
         <item>How do I apply for new loan account?</item>
@@ -469,7 +471,7 @@
     <string name="pref_base_url_title">Update Endpoint</string>
     <string name="pref_base_url_desc">Click here to change your Endpoint Configurations</string>
     <string name="pref_configuration">pref_configuration</string>
-    <string name="enter_base_url">Enter the Base Url</string>
+    <string name="enter_base_url">Enter the Base URL</string>
     <string name="enter_tenant">Enter the Tenant</string>
     <string name="account_type_loan" translatable="false">accountType.loan</string>
     <string-array name="languages" translatable="false">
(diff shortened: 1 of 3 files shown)


miPlodder:
@therajanmaurya @dilpreet96 @satyan @droidchef Please review this PR.

miPlodder:
@therajanmaurya @dilpreet96 @satyan @droidchef Please review this PR.

app/src/main/res/values/strings.xml
```diff
@@ -383,6 +383,8 @@
     <string name="language">Language</string>
     <string name="choose_language">Choose your language</string>
     <string name="notification">Notifications</string>
+    <string name="base_url">Base Url</string>
```
luckyman20: It should be `Base URL`

luckyman20:
@miPlodder Another approach you can use that check all the things when `OK` button is clicked, it will shorten the code.

You should also check whether it is a valid URL or not

app/src/main/res/values/strings.xml
miPlodder: @luckyman20 I have used "Base Url" because the Title also shows "Base Url". It's better to have consistency in messages.

miPlodder:
@luckyman20  I have made a check for validity of the URL. Moreover, when `OK` button is pressed, then the AlertDialog closes, we cannot prevent this behaviour, please refer this `abstract class PreferenceDialogFragmentCompat`, where it creates iAlert dialog and we are using it.

Moreover, it's good approach to give continuous feedback to user in forms where lesser number of fields are available. That's why I have used TextWatcher.

app/src/main/res/values/strings.xml
luckyman20: @miPlodder I think you can change that too also because URL is a short form for Uniform Resource Locater & short forms should be written in caps only.

therajanmaurya:
Fix conflicts.

miPlodder:
@therajanmaurya Please review this PR again.

PR #824, opened 2018-05-07
Fix #814 : Added proper descriptive message during Transfer process in all use-cases

### LINKED ISSUE #814: Should display more descriptive error message when not enough funds in chosen saving account to repay loan
**Summary:**

When attempting to repay a loan from a savings account that doesn't have enough funds, error message isn't as accurate as it should be

**Steps to reproduce:**

Navigate to an active loan, re pay it but sleeve an amount greater than the balance in the chosen savings account. 

**Expected behaviour:**

Error message should be specific and say there is not enough funds, not the vague error message it states ( see evening

Android 6.0.1
screenshot_20180421-012149 up

**Screenshots:**

Can be created by pressing the Volume Down and Power Button at the same time on Android 4.0 and higher.

miPlodder:
Fixes #814 

pr_unsufficient_balance


miPlodder:
@therajanmaurya Please review this PR.

PR #855, opened 2018-06-03
Feat #840 : Added Support for Reports [with demo API]

### LINKED ISSUE #840: Support for viewing Reports
**Details :**

We can enable the user to view reports and even download the report to their mobile devices. We can even add an option to send report as an email to the Email Id of the user.

miPlodder:
Fixes #840 

pr_840


app/src/main/java/org/mifos/mobilebanking/utils/CheckSelfPermissionAndRequest.java
```diff
@@ -141,4 +142,81 @@ public void onClick(DialogInterface dialog, int which) {
             }
         }
     }
+
+
+    @TargetApi(Build.VERSION_CODES.JELLY_BEAN)
+    public static void requestPermissionFragment(final Fragment fragment,
```
miPlodder: @therajanmaurya Please check this approach and suggest alternatives/improvements.
I have added this so that I can get the PermissionResult Callback in fragment, other method was for Activity.

miPlodder:
@therajanmaurya Currently, I have not shown anything on Viewing Screen as the actual format .pdf or .xls is not known.

prashantkh19:
@therajanmaurya We will be working on this issue during this GSoC and my approach is a bit different than this. We would discuss the approach at that time. I think we can leave this PR as of now.

garvit984:
@miPlodder What should we do about this PR?

miPlodder:
@garvit984 I see this functionality not being implemented yet wrt backend. Let's keep it for reference purposes for future work wrt this functionality.

@edcable We can add it as a Task for this year GSoC if it's still a requirement.

miPlodder:
Changes look depreciated.
@edcable We add it as a use-case for this year's GSoC.

PR #876, opened 2018-08-29
Fix #847: Username and Image should be shown when Internet is not there in Home Fragment.

### LINKED ISSUE #847: Username and Image should be shown when Internet is not there in Home Fragment
**Details :**

Currently, what happens is that in Home Fragment, Username’s TextView is blank, because it’s
being fetched for some time. Similarly, Total Savings and Total Loan Amount fields are also
empty. It’s a bad UI design in case of no or slow Internet.
Instead, we can store them in the Shared Preference and show them when Internet Connection is
slow, and update the Shared Prefs and data on the fields when the response from the Server is
fetched.

_image

We can atleast store and use Username, and Account Number for the Logged In user. We can
use Shared Preference for storing and clean them when the User logs out.
Moreover, ImageView shows “M” that means it uses Mifos as username when username is not
fetched. Instead we can use first character of the user name (that we have stored in Shared
Preference) and display it. And update the ImageView when the image is fetched. Even we can
cache the image and display it on ImageView.

jaindiv26:
Description:

I've stored the entered username while login in the prefrenceHelper.java and set the navigational username initially from prefrenceHelper, set the totalLoanAmount and totalSavingsAmount to "N/A" just before the API call and so in case of slow internet user won't see empty textView and the shared prefs and data on the fields will update, when the response from the Server is fetched.

#Issue_Number : #847 
Username and Image should be shown when the Internet is not there in Home Fragment.

Screenshots.

screenshot_1535541506

screenshot_1535541502

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/activities/LoginActivity.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/activities/LoginActivity.java
@@ -9,6 +9,7 @@ import android.widget.LinearLayout;
 import android.widget.Toast;
 
 import org.mifos.mobilebanking.R;
+import org.mifos.mobilebanking.api.local.PreferencesHelper;
 import org.mifos.mobilebanking.presenters.LoginPresenter;
 import org.mifos.mobilebanking.ui.activities.base.BaseActivity;
 import org.mifos.mobilebanking.ui.views.LoginView;
@@ -63,6 +64,8 @@ public class LoginActivity extends BaseActivity implements LoginView {
     @Override
     public void onLoginSuccess(String userName) {
         this.userName = userName;
+        PreferencesHelper preferencesHelper = new PreferencesHelper(this);
+        preferencesHelper.setUserName(userName);
         loginPresenter.loadClient();
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mifos/mobilebanking/ui/fragments/HomeOldFragment.java
```diff
@@ -205,6 +205,8 @@ public void onRefresh() {
     }
 
     private void loadClientData() {
+        tvSavingTotalAmount.setText("N/A");
```
miPlodder, line 208: @jaindiv26 Don't set it to `NA` which means something else, either error message should be consistent or try a hack. The hack here will be to store total saving and total loan amount in shared preference.

miPlodder (changes requested):

app/src/main/java/org/mifos/mobilebanking/ui/activities/LoginActivity.java
```diff
@@ -63,6 +64,8 @@ protected void onCreate(Bundle savedInstanceState) {
     @Override
     public void onLoginSuccess(String userName) {
         this.userName = userName;
+        PreferencesHelper preferencesHelper = new PreferencesHelper(this);
```
miPlodder, line 67: @jaindiv26 Don't create Instance like this, use Dagger Annotation Inject to do the same.

miPlodder (changes requested):

PR #877, opened 2018-08-30
Fix #845: Make Writing and Displaying larger number convenient for User.

### LINKED ISSUE #845: Make Writing and Displaying larger number convenient for User
**Details :**

For instance, while Signing Up, user has to enter his/her Account Number, it’s difficult for the
user to count the number of zeroes entered or even validate, the entered account number
properly.
A better approach will be to automatically add spaces after every 3 digits entered for better
visibility to the user. It can also be used while entering his/her phone number (even on helpline
phone number), or any place where user is asked to enter large number/text This approach helps
user to visualize the long number in better way. Somewhat like this

(For Account Number)
image

(For Phone Number)
image

jaindiv26:
Description: 
I've changed the input type from number to phone in the XML, and in sign-up activity whenever the user enters its account number or mobile number the spaces will be included, even when the user edits something in the field the spaces will still be there, I've covered all the cases.

Fixes #845 
Make Writing and Displaying larger number convenient for User.

Screenshots:
screenshot_1535651457

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
@@ -2,6 +2,8 @@ package org.mifos.mobilebanking.ui.fragments;
 
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.text.Editable;
+import android.text.TextWatcher;
 import android.util.Patterns;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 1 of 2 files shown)


luckyman20:
@jaindiv26 Please fix the Travis Error

luckyman20:
@jaindiv26 Please squash all the commits in one

miPlodder:
@jaindiv26 Have you covered the case if user erases a digit or the space that you have added and what if we fetch the number from EditText, it should not include the spaces. Few more case may arise.

miPlodder:
@jaindiv26 Good job, but I'm working on a more generic way so that thay can be applied to all the numeric or say alphanumeric EditTexts.

You can work on other issues.

PR #878, opened 2018-09-02
Fix #801: Adding password strength checker while registering a new client.

### LINKED ISSUE #801: feat: adding password strength checker while registering a new client.
**Summary:**

Currently on signing up the strength of password while setting it up is not shown.Since this is a banking app I think it would be pretty helpful if we suggest the strength of the password so that user can set up a strong password for better security.
I want to add something like this:
4xh8y

jaindiv26:
Description: 
I've added a password strength checker while registering a new client. Which will tell the user whether the entered password is weak, medium, strong or a very strong type!

Fixes #801 
Adding a password strength checker while registering a new client.

Screenshots: 
screenshot_1535909886
screenshot_1535909910
screenshot_1535909919
screenshot_1535910318


miPlodder:
@jaindiv26 Squash the commits and fix the Travis Build.

therajanmaurya:
@jaindiv26 Please fix the Travis and squash your commits

therajanmaurya:
Fixed in #1173

PR #879, opened 2018-09-03
Fix #782: Adding check for network connection in activities and fragments.

### LINKED ISSUE #782: Adding check for network connection in activities and fragments.
**Summary:**

In many fragments, while showing error, network connectivity is not checked which leads to a less suitable error message like "error loading in {Feature name}" whereas the message should reflect that there is no internet connection.

**Screenshots:**

screenshot_20180323-070210

jaindiv26:
Description: 
Adding check for network connection in activities and fragments.

Fixes #782 
Adding check for network connection in activities and fragments.

Screenshots: 
screenshot_1535994195

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/utils/Network.java
+++ b/app/src/main/java/org/mifos/mobilebanking/utils/Network.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.telephony.TelephonyManager;
+import android.util.Log;
 
 /**
  * Created by rishabhkhanna on 07/03/17.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mifos/mobilebanking/utils/Network.java
```diff
@@ -22,6 +23,30 @@ public static NetworkInfo getNetworkInfo(Context context) {
...
+     */
+
+    public static boolean isInternetOn(Context context) {
+        ConnectivityManager cm = (ConnectivityManager) context
+                .getSystemService(Context.CONNECTIVITY_SERVICE);
+        // test for connection
+        if (cm.getActiveNetworkInfo() != null
+                && cm.getActiveNetworkInfo().isAvailable()
+                && cm.getActiveNetworkInfo().isConnected()) {
+            Log.v("log", "Internet is working");
```
luckyman20, line 40: Log commands should be removed

app/src/main/java/org/mifos/mobilebanking/utils/Network.java
```diff
@@ -22,6 +23,30 @@ public static NetworkInfo getNetworkInfo(Context context) {
...
+            // txt_status.setText("Internet is working");
```
luckyman20, line 41: Remove this comment also

ivary43:
@jaindiv26  fix the travis CI issue.

therajanmaurya:
First Implementation is totally wrong. You have added extra code that is already present in utils classes. Solution is to check internet connection from util class and show no network UI of sweet error.

PR #904, opened 2018-10-24
Switched order of jcenter() / google() references

etlhsu:
Switching the order of the references helps prevent Gradle build issues when trying to reference certain support libraries.


etlhsu:
The modification will be included as a part of #905

Issue #896, opened 2018-10-20
Error text beneath EditText should be shown for all the EditText on clicking the Sign Up Button


miPlodder:
**Summary:**

Error text beneath EditText should be shown for all the EditText on clicking the Sign Up Button

**Steps to reproduce:**

Sign Up without entering any details

**Expected behaviour:**

Errors for all the EditText should be shown like shown in other apps like Paypal

**Observed behaviour:**

Error text are not shown for all the EditText

**Device and Android version:**

Xiaomi, Marshmallow

**Screenshots:**

![20181020_122528](https://user-images.githubusercontent.com/22195621/47252635-96b48b00-d465-11e8-8ba8-911709489845.gif)

letelete:
I'll take this issue

letelete:
Review ready! https://github.com/openMF/mifos-mobile/pull/907

shiv07tiwari:
@miPlodder Can I work on this issue.
I am interested in contributing to this organisation .

shiv07tiwari:
@letelete  Hello. Can I take up this issue or this is still under process ?

PR #907, opened 2018-10-24
Feat: #896 Add error message for all fields in RegistrationFragment

letelete:
Add error message for all fields.

# Input error preview
input_error_presentation

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
@@ -1,11 +1,15 @@
 package org.mifos.mobilebanking.ui.fragments;
 
+import android.content.Context;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.design.widget.TextInputLayout;
+import android.text.TextUtils;
 import android.util.Patterns;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.RadioButton;
 import android.widget.RadioGroup;
@@ -24,6 +28,8 @@ import javax.inject.Inject;
 import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
+import butterknife.OnFocusChange;
+import butterknife.OnTextChanged;
 
 /**
  * Created by dilpreet on 31/7/17.
@@ -61,7 +67,12 @@ public class RegistrationFragment extends BaseFragment implements RegistrationVi
     @Inject
     RegistrationPresenter presenter;
 
+    private static final String LIVE_UPDATE_ENABLED_TAG = "live_update_enabled";
+    private static final String LIVE_UPDATE_DISABLED_TAG = "live_update_disabled";
+
     private View rootView;
+    private boolean areAllInputsCorrect;
+    private EditText firstInvalidInput;
 
     public static RegistrationFragment newInstance() {
         RegistrationFragment fragment = new RegistrationFragment();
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -19,8 +19,8 @@ buildscript {
 
 allprojects {
     repositories {
-        jcenter()
         google()
```
miPlodder: @letelete Undo this change

app/src/main/java/org/mifos/mobilebanking/ui/views/RegistrationView.java
```diff
@@ -1,5 +1,7 @@
 package org.mifos.mobilebanking.ui.views;
 
+import android.widget.EditText;
```
miPlodder: @letelete Undo this change

miPlodder:
@letelete Fix the Travis Build failed and add proper GIF showing UI changes.
Moreover, instead of creating multiple commits for small functionality change, create a single commit for the issue.

letelete:
Is there any way to edit these, already existing commits and squash them all into one? I'm not sure how can I could've done that. Thought it's being done by a person who's merging the PR.

letelete:
@miPlodder travis failed with `Execution failed for task ':app:getDependencies'.`, which I've struggled with when first time cloning a repo. I fixed it on my machine by moving google() before jcenter() in build.gradle

build.gradle
etlhsu: Pull request #909 covers this change, and a fix for multidex

miPlodder:
@letelete You can use this [link](http://www.jancarloviray.com/blog/how-to-merge-multiple-commits-into-one-git-commit/) to squash the commit to one. Its job of the developer to maintain a clean Commit history.

> @miPlodder Travis failed with `Execution failed for task ':app:getDependencies'.`, which I've struggled with when first time cloning a repo. I fixed it on my machine by moving google() before jcenter() in build.gradle

Create an issue for Travis Build failed to explain what's happening. I will look into it.

app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
```diff
@@ -115,61 +151,153 @@ public void registerClicked() {
     }
 
     private boolean areFieldsValidated() {
```
miPlodder: @letelete What you are doing is correct. But instead of creating methods for every `if`, write them in a single method like it was written before.

letelete: @miPlodder **Isn't it a bad practice to create long methods?** 

It's so unreadable and hard to maintain if the single method contains all of the validation logic. It's also easy to repeat yourself here, and that's what I've seen in your code before my fix. There were 2 validations for some of the input (I guess it was a password) which was checking if password length is greater than a) 5, b) 6 which is complete nonsense and it 100% comes out of hard-to-read method with tons of validations.

I've also seen validations for some specific inputs which were not grouped-up so at the beginning I could spot X validation and at the end of the method, I could also spot validation for the same X input.

**My solution ensures clean-code standards.** Methods are easy to maintain now and if we want to change some of the validation we just need to search for a specific method instead of analyzing one, enormous method.

**Splitting validations into method has also another positive aspect**, which is a fact, that with the live-update feature, we don't need to check every single validation and just the one we are interested in(validation for currently edited input).

letelete:
I've added an issue describing the build problem #910

letelete:
Fixing the bug described in #910 makes build successful. I thought you would like me to remove gradle.build change since It's not the PR for that.

app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
miPlodder: @letelete I know that your code is cleaner. But I have a better alternative which I will add to the GCI tasks in future. 
We can create a Utils Class where we can add all the validations methods (like you added here). By that way, we can call those methods, for any of the EditText (like Sign up, Login) where that validation is required. This will reduce redundant code for validation.

letelete: Yeah, that's actually a good idea which I was thinking about when faced with writing the same code for login_activity validation(#911) 😆 I will get all of these and split them into one as it was before, then.

letelete:
@miPlodder Done. Btw. I've removed live update feature for now and will add it again when Utill class is ready.

app/src/main/java/org/mifos/mobilebanking/ui/fragments/RegistrationFragment.java
```diff
@@ -96,12 +99,6 @@ public void registerClicked() {
             payload.setFirstName(etFirstName.getText().toString());
             payload.setLastName(etLastName.getText().toString());
             payload.setMobileNumber(etPhoneNumber.getText().toString());
-            if (!etPassword.getText().toString().equals(etConfirmPassword.getText().toString())) {
```
miPlodder, line 99: @letelete Good fix here. :+1:

letelete:
@miPlodder Add request view feature which I told you about earlier. Speaking about live update validation. I would like to add it later when we will create separate validation class which you were talking about.

# Preview

preview

letelete:
### Add new validation logic

* If the input has focus for the first time:
  * don't use live-update validation
* If user left inputX (and move to inputY): 
  * check correctness of inputX 
    * if invalid - show error
* If user back to the previous input (in this case: from inputY to inputX) **AND** input is invalid:
  * turn on live-update
* if user back to the previous **VALID** input
  * don't use live-update validation
* On register button click:
  * show error for every invalid input
  * turn on live-update validation for every input
  * move to the first invalid input and request focus on it
  * force keyboard

miPlodder:
@letelete You logic is inconsistent. Try different combination on the device. You can work on this in your spare time. You can focus on newer tasks.

miPlodder:
@letelete Good job, but we are coming up with new design for Sign-Up in which we are splitting the fields in different fragments which makes this PR obsolete. Anyways good job here.

Issue #895, opened 2018-10-20
Error text beneath EditText should be shown for all the EditText on clicking the Login Button


miPlodder:
**Summary:**

Error text beneath EditText should be shown for all the EditText on clicking the Login Button

**Steps to reproduce:**

Log into app without entering credentials

**Expected behaviour:**

Errors for all the EditText should be shown like shown in other apps like Paypal

**Observed behaviour:**

Error text are not shown for all the EditText

**Device and Android version:**

Xiaomi, Marshmallow

**Screenshots:**

![20181020_122512](https://user-images.githubusercontent.com/22195621/47252571-923ba280-d464-11e8-8f28-192c3e94d9f9.gif)

letelete:
I'm working on it.

letelete:
Waiting for review. BTW, I couldn't find this task on the GCI site, why is that?

Issue #910, opened 2018-10-25
Bug: Unresolved dependencies error while building app


letelete:
**Summary:**

Can't build the app, because of compiler unresolved dependencies errors, such as:

> Failed to resolve: multidex
> Failed to resolve: multidex-instrumentation

**Steps to reproduce:** 

1. Clone development branch
2. Open project in your IDE
3. Build project

**Expected behaviour:**

Complete build

**Observed behaviour:**

As I mentioned before, there are unresolved dependencies errors:

![image](https://user-images.githubusercontent.com/18665370/47484032-f7174400-d83a-11e8-9c3b-0e94987db872.png)

Running `./gradlew --debug check` logs:
```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':app:getDependencies'.
Could not find multidex.jar 
```
For more details, check pastebin where I've posted more details from logs
 https://pastebin.com/XiXbYFAW

**Device and Android version:**

Mobile device doesn't matter here, so I'll just say that I'm using Android Studio as an IDE

**Screenshots:**

![image](https://user-images.githubusercontent.com/18665370/47484124-42c9ed80-d83b-11e8-9905-b47ad510db7f.png)

letelete:
@miPlodder  This issue appears in #907, #908. Fix for this issue might be #909

migueldinisdev:
I was having this issue in my project after cloning it. I actually solved it by changing some stuff in **build.gradle**. I added it to **.gitignore** though.

UPDATE:

What I did was a temporary workaround in my computer. I increased **minSdkVersion** to **21** and removed multidex library dependency.

As described [here](https://developer.android.com/studio/build/multidex), projects with minSdkVersion >= 21 do NOT require multidex library to be imported in build.gradle. That way I avoided that error.

_This is only a temporary solution working as a workaround_

m-sameer:
@miPlodder I'm able to fix this issue by updating the `compileSdkVersion` and some dependencies. Can you please add this as a GCI task?

PR #911, opened 2018-10-25
Feat: #895 Add error message for all the EditText in LoginActivity

letelete:
### Info
The error is shown for both inputs at once now. 

I didn't create live-update function here, because I think is redundant in login since user should know, his correctly set credentials, so error on button click does the job 

**I couldn't find this task on GCI site tho, why is that?**

### Preview

preview


letelete:
Fixing the bug described in #910 makes build successful. I thought you would like me to remove gradle.build change since It's not the PR for that.

letelete:
Squash every validation into one method for now, since there will be some Utill class to improve that.

PR #909, opened 2018-10-24
Fix: Improved multidex and repository declaration

etlhsu:
Fix #910: Unresolved dependencies error while building app

The commit fixes issues where certain references caused issues when building project

PR #913, opened 2018-10-25
Fix #895: Add error text below password EditText in LoginActivity

### LINKED ISSUE #895: Error text beneath EditText should be shown for all the EditText on clicking the Login Button

m-sameer:
Fixes #895 

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
+++ b/app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
@@ -175,36 +175,42 @@ public class LoginPresenter extends BasePresenter<LoginView> {
 
 
     private boolean isCredentialsValid(final String username, final String password) {
-
+        boolean credentialValid = true;
         final Resources resources = context.getResources();
         final String correctUsername = username.replaceFirst("\\s++$", "").trim();
         if (username == null || username.matches("\\s*") || username.isEmpty()) {
             getMvpView().showUsernameError(context.getString(R.string.error_validation_blank,
                     context.getString(R.string.username)));
-            return false;
+            credentialValid = false;
         } else if (username.length() < 5) {
             getMvpView().showUsernameError(context.getString(R.string.error_validation_minimum_chars
                     , resources.getString(R.string.username), resources.getInteger(R.integer.
                             username_minimum_length)));
-            return false;
+            credentialValid = false;
         } else if (correctUsername.contains(" ")) {
             getMvpView().showUsernameError(context.getString(
                     R.string.error_validation_cannot_contain_spaces,
                     resources.getString(R.string.username),
                     context.getString(R.string.not_contain_username)));
-            return false;
-        } else if (password == null || password.isEmpty()) {
+            credentialValid = false;
+        } else {
+            getMvpView().clearUsernameError();
+        }
+
+        if (password == null || password.isEmpty()) {
             getMvpView().showPasswordError(context.getString(R.string.error_validation_blank,
                     context.getString(R.string.password)));
-            return false;
+            credentialValid = false;
         } else if (password.length() < 6) {
             getMvpView().showPasswordError(context.getString(R.string.error_validation_minimum_chars
                     , resources.getString(R.string.password), resources.getInteger(R.integer.
                             password_minimum_length)));
-            return false;
+            credentialValid = false;
+        } else {
+            getMvpView().clearPasswordError();
         }
 
-        return true;
+        return credentialValid;
     }
 
     /**
--- a/app/src/main/java/org/mifos/mobilebanking/ui/activities/LoginActivity.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/activities/LoginActivity.java
@@ -111,6 +111,16 @@ public class LoginActivity extends BaseActivity implements LoginView {
         tilPassword.setError(error);
     }
 
+    @Override
+    public void clearUsernameError() {
+        tilUsername.setErrorEnabled(false);
+    }
+
+    @Override
+    public void clearPasswordError() {
+        tilPassword.setErrorEnabled(false);
+    }
+
     /**
      * Called when Login Button is clicked, used for logging in the user
      */
@@ -119,8 +129,6 @@ public class LoginActivity extends BaseActivity implements LoginView {
 
         final String username = tilUsername.getEditText().getEditableText().toString();
         final String password = tilPassword.getEditText().getEditableText().toString();
-        tilUsername.setErrorEnabled(false);
-        tilPassword.setErrorEnabled(false);
 
         if (Network.isConnected(this)) {
             loginPresenter.login(username, password);
--- a/app/src/main/java/org/mifos/mobilebanking/ui/views/LoginView.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/views/LoginView.java
@@ -38,6 +38,9 @@ public interface LoginView extends MVPView {
 
     void showPasswordError(String error);
 
+    void clearUsernameError();
+
+    void clearPasswordError();
 
     /**
      * Should be called when the client is fetched successfully from API.


luckyman20:
@m-sameer Fix the Travis Build Error

m-sameer:
@luckyman20 Done. :smile:

app/build.gradle
```diff
@@ -77,6 +77,9 @@ dependencies {
     implementation "com.google.android.gms:play-services-oss-licenses:$rootProject.playServicesVersion"
     implementation "com.isseiaoki:simplecropview:$rootProject.cropviewVersion"
 
+    // Multidex
```
miPlodder: @m-sameer Remove it. PR #909  will be merged then you can pull that code and update your PR.

build.gradle
```diff
@@ -19,8 +19,8 @@ buildscript {
 
 allprojects {
     repositories {
-        jcenter()
         google()
+        jcenter()
```
miPlodder: Remove this too.

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
```diff
@@ -178,22 +178,25 @@ private boolean isCredentialsValid(final String username, final String password)
 
         final Resources resources = context.getResources();
         final String correctUsername = username.replaceFirst("\\s++$", "").trim();
+        boolean usernameValid = true;
```
miPlodder: @m-sameer Check your logic, no additional variable is required and no checks for passwords are added.

m-sameer: @miPlodder If I don't use an additional variable for checking the username validity and return false directly the password **EditText** will not be checked. :confused: 
These are the minimal changes I can think of, can you suggest any?

miPlodder: @m-sameer Then, don't return at that stage. Try to improve your solution and use a `TextWatcher` then you wouldn't be needing any additional variables.

m-sameer: @miPlodder If I use a `TextWatcher` the credentials will be checked on every tap on the keyboard but they should be checked only when the login button is pressed. Also adding a listener will be more complicated and will add more lines than simply using a boolean variable.

miPlodder: @m-sameer It will give a better UX. Try to sign up/ sign in to any software, which one will you be convenient, the one which you get an instant feedback.

I know it will be more work. You don't worry about the deadline, I will extend it and don't focus on doing a high number of tasks, but focus on doing Quality tasks.

When free read [this](http://www.howdesign.com/web-design-resources-technology/donald-normans-design-principles-modern-web-design/).

luckyman20: @m-sameer @miPlodder I have another approach which might satisfy need of both of you.

You can add a `onFocus` of Password EditText and then check if username is null or not.

miPlodder: @luckyman20 That is what I'm talking. For example, if a user is entering his username, we don't want the validation/textwatcher to work. When he changes his focus to another field, then the textwatcher should be active. Now, when user edits his username, how the textwatcher will be active and showing all the relevant errors. 

See Paypal mobile interface while signing up.

luckyman20: @miPlodder I think we won't be needing a TextWatcher for this, we have to split the `isCredentialsValid` into two separate methods namely `isUsernameValid` and `isPasswordValid` and we will call `isUsernameValid` in onFocus of Password EditText and then call `isPasswordValid` in `onSubmit`.

m-sameer: @miPlodder I've added an `OnFocusChangeListener` which checks the credentials on focus change. Earlier I was checking the credentials on button click because the issue said **Error text beneath EditText should be shown for all the EditText on clicking the Login Button**.
I agree that we should strive to improve the UX and the link to the design website was helpful. :)
As you said, I'm focused on doing high-quality tasks rather than higher quantity.

@miPlodder @luckyman20 Thanks for your help.

m-sameer:
@miPlodder Please review.

luckyman20:
@m-sameer Fix the travis build

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
```diff
@@ -75,7 +75,7 @@ public void detachView() {
      */
     public void login(final String username, final String password) {
         checkViewAttached();
-        if (isCredentialsValid(username, password)) {
+        if (isUsernameValid(username) & isPasswordValid(password)) {
```
m-sameer: The build is failing because I'm using `&` here instead of `&&`. I cannot use **short circuit and** operation because both sides of the condition need to be called to show error on the username as well as the password field. @miPlodder As you said I'm not using any extra variables but this is failing on the **findbugs** check with the warning **NS_DANGEROUS_NON_SHORT_CIRCUIT: Potentially dangerous use of non-short-circuit logic**.

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
```diff
@@ -75,7 +75,7 @@ public void detachView() {
      */
     public void login(final String username, final String password) {
         checkViewAttached();
-        if (isCredentialsValid(username, password)) {
+        if ((isUsernameValid(username) | isPasswordValid(password)) == 0) {
```
miPlodder: Use boolean functions and why are you doing Bitwise OR.

miPlodder (changes requested):

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
m-sameer: As I have mentioned before, if I use **boolean** and use logical `&&` `isPasswordValid` will not be called if `isUsernameValid` return false as a result the error for the password field will not be shown.

Also, If I'm using **boolean** and bitwise `&` the build is failing with the error **NS_DANGEROUS_NON_SHORT_CIRCUIT: Potentially dangerous use of non-short-circuit logic** as we should not use bitwise operators for **boolean** values.

Bitwise operator is used to call both the methods in any case to show the error for both the fields (username and password).

miPlodder: Yeah, good hack. But this is not a good way to do it. That is what Build is also trying to say. Try to see #907 and implement it. (not fully, but partially as that code has more features also)

Separating the methods for validation is not needed.

m-sameer: @miPlodder That will require a `boolean` variable as I was doing it before. :confused:

m-sameer:
@miPlodder I've updated the PR again using boolean variables and the approach is similar to #907. I've tried using many approaches, if anything is wrong with this one please let me know. The build is passing and there are no performance issues.

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
```diff
@@ -175,25 +175,27 @@ public void onNext(Page<Client> clientPage) {
...
+            credentialValid = false;
         } else if (correctUsername.contains(" ")) {
             getMvpView().showUsernameError(context.getString(
                     R.string.error_validation_cannot_contain_spaces,
                     resources.getString(R.string.username),
                     context.getString(R.string.not_contain_username)));
-            return false;
-        } else if (password == null || password.isEmpty()) {
+            credentialValid = false;
+        }
```
miPlodder, line 198: @m-sameer Add else and hide errors there. See #907 for reference.

miPlodder (changes requested):

miPlodder:
@m-sameer And update the GIF after changing the solution.

m-sameer:
The errors are cleared on the login button click.
```
tilUsername.setErrorEnabled(false);
tilPassword.setErrorEnabled(false);
```

The GIF is already up to date.

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
miPlodder: @m-sameer Remove [this](https://github.com/openMF/mifos-mobile/blob/development/app/src/main/java/org/mifos/mobilebanking/ui/activities/LoginActivity.java#L122)
and add else there remove the error.

app/src/main/java/org/mifos/mobilebanking/presenters/LoginPresenter.java
```diff
@@ -175,25 +175,27 @@ public void onNext(Page<Client> clientPage) {
...
+
+        if (password == null || password.isEmpty()) {
             getMvpView().showPasswordError(context.getString(R.string.error_validation_blank,
                     context.getString(R.string.password)));
             return false;
```
miPlodder: @m-sameer Don't return the value here. Instead, follow a similar approach that you did for Username

miPlodder (changes requested):

m-sameer:
@miPlodder The requested changes have been incorporated.

miPlodder:
@m-sameer Good job

m-sameer:
@miPlodder Thank you!

PR #921, opened 2018-11-01
Fix #920: Splash Screen Image Streching

### LINKED ISSUE #920: Splash Screen Image Streching
**Summary:**

When the splash screen is opened, a stretched version of the Mifos logo appears, making the non-circular globe and lettering look odd.

**Steps to reproduce:**
1. Close the app, make sure all activities have been destroyed and the UI is dismissed from the background memory.
2. Open the app and observe the splash screen.
3. Capture the splash screen using a well-timed screenshot or video recording.

**Expected behaviour:**

A non-stretched form the Mifos logo should appear. The image should scale on both axises, to prevent distorting the content.

**Observed behaviour:**

A horizontally stretched form the Mifos logo appears, as shown by the non-circular glove and wider-than-usual lettering.

Google Pixel, Physical Device and Emulator. Android 9.0 Pie.

**Screenshots:**

image

etlhsu:
Fixes #920 

The splash screen now includes an untampered version of the Mifos logo. To achieve such effect, the height attribute was removed, so the height of the image could expand to match the original image ratio. 

image


luckyman20:
@ethan627hsu Can you please upload a screenshot that how did it looked before your changes?

miPlodder:
@ethan627hsu I don't think this as a generic solution, the hardcoded value works for your device, not all devices. 

Instead, a better solution for this issue is to add a SVG image as a logo, which will make it stretching of the image independent of device size.

etlhsu:
@miPlodder, I have added a non-hard coded solution, that scales to different device sizes. After this pull request experience, I have now learned to check my solution to make sure it fits corner cases. I am hoping to provide better solutions in the future, that account for different configurations. Thanks for the suggestion!

etlhsu:
@luckyman20, you can view the original splash screen by visiting the #920 issue page.

app/src/main/res/drawable/splash_background.xml
```diff
@@ -4,9 +4,7 @@
     <item android:drawable="@color/white"/>
 
     <item
-        android:height="100dp"
-        android:width="350dp"
-        android:drawable="@drawable/mifos_logo"
+        android:drawable="@drawable/splash_mifos_logo"
```
miPlodder, line 7: @ethan627hsu Rename these files to `mifos_logo`

miPlodder:
@ethan627hsu The solution is correct, but the image over-sizes on my device. From where have you taken these images?

I think the images are not in the standard sizes.

therajanmaurya:
@ethan627hsu Please make the suggested changes.

therajanmaurya:
yes these are not in standard sizes. So I don't think so we can merge and last PR inactive from long time.

PR #922, opened 2018-11-01
Fix #919: Adapter Attachment Error in RecyclerView when viewing "Charges" and "Accounts" pages

### LINKED ISSUE #919: Adapter Attachment Error in RecyclerView when viewing "Charges" and "Accounts" pages
**Summary:**

When opening the main pages for viewing accounts and charges, an error log is written speaking of how a recyclerview handled a case where no adapter is attached to it. I decided to have this issue scope both "Accounts" and "Charges" pages as the exact same log is sent out with a very similar entry point.  

**Steps to reproduce:**

1. Run the Mifos Mobile app through an IDE, and validate that the device is able to log information through Logcat.
2. While in the Logcat view, filter the logs to tag `E/RecyclerView` by typing the tag into the search bar, this will help show the error log clearly rather than combined with other log messages.
2. Go through the splash screen with the password login.
3. At the main home screen, click on the "Charges" or "Accounts" button.
4. Observe the new error log that has been added to Logcat.

**Expected behaviour:**

The app should not be reporting this kind of error, adapters should be attached to their respective RecyclerViews, and RecyclerViews with no adapter should be dealt with in a way that causes no error. 

**Observed behaviour:**

Adapters from the "Charges" and "Accounts" pages are not being attached to respective RecyclerViews when opened, the RecyclerViews log an error message signifying the behavior:

`E/RecyclerView: No adapter attached; skipping layout`

Google Pixel, Physical Device and Emulator. Android 9.0 Pie.

etlhsu:
Fixes #919 

Changes improve the declaration and initialization of adapters, and binding data to them properly. These efforts eliminate a non-fatal error log, stating how RecyclerViews were being displayed with no adapter connected to them.


app/src/main/java/org/mifos/mobilebanking/ui/fragments/AccountsFragment.java
```diff
@@ -128,6 +128,7 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
         accountsPresenter.attachView(this);
         sweetUIErrorHandler = new SweetUIErrorHandler(getActivity(), rootView);
 
+        rvAccounts.setVisibility(View.GONE);
         LinearLayoutManager layoutManager = new LinearLayoutManager(getActivity());
         layoutManager.setOrientation(LinearLayoutManager.VERTICAL);
         rvAccounts.setLayoutManager(layoutManager);
```
miPlodder, line 134: @ethan627hsu  Changing visibility is not an appropriate solution. Instead, add `adapter` to `recyclerview` inside `onCreateView()` method.

miPlodder (changes requested):

miPlodder:
@ethan627hsu Make the mentioned changes for both the fragments.

therajanmaurya:
@ethan627hsu @miPlodder is right. this is not the solution. Let me tell you first when we get 
`E/RecyclerView: No adapter attached; skipping layout`

As we inflate View in `OnCreateView` or `OnCreate`. It inflate the views on screen and when we does not set recycler view adapter then it gives warning, the above warning `E/RecyclerView: No adapter attached; skipping layout`. 

The solution for this warning is to set the adapter with empty list and later when you get the data notify recyclerview about the new data by notifying adapter `adapter.notifyDataSetChanged()`

Make the changes and let me know.

therajanmaurya:
InActive Closing It.

PR #939, opened 2018-11-19
Fix #932: Snackbar pops on the FAB on deleting Beneficiary

### LINKED ISSUE #932: Snackbar pops on the FAB on deleting Beneficiary
**Summary:**

Snackbar pops on the FAB on deleting Beneficiary 

**Steps to reproduce:**

Beneficiary > Click on any Beneficiary or create new Beneficiary > Delete it from menu

Create new Beneficiary using details,

> Account Number = 000000001
> Office Number = Head Office
> 
> Rest information can be anything.
> 

**Expected behaviour:**

FAB should animate upwards so that it doesn't get covered by Snackbar 

**Observed behaviour:**

Instead, FAB doesn't animate.

Xiaomi, Marshmallow

**Screenshots:**

etlhsu:
Fixes #932 

The bug was caused by the snackbar being a part of the wrong layout. Previously, the snackbar had been set to appear in the layout of a BeneficiaryDetailFragment. The snackbar would appear in the BeneficiaryDetailFragment, but the fragment would be quickly destroyed, as the work is done when the fragment is being destroyed. This leaves the snackbar to just "be there" and appear above any other UI elements.

My solution moves the snackbar to be a part of the BeneficiaryListFragment layout, the screen that lists out the beneficiaries and has the FAB. By doing this, the FAB is able to be aware of the snackbar and naturally moves up. I also implemented the same functionality for the "update beneficiary" action, as my solution scales to fix other cases of the issue. No more left out snackbars! 🍫🍬
  
image

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryApplicationFragment.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryApplicationFragment.java
@@ -3,6 +3,7 @@ package org.mifos.mobilebanking.ui.fragments;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.support.design.widget.TextInputLayout;
+import android.support.v4.app.Fragment;
 import android.support.v4.widget.NestedScrollView;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -286,10 +287,12 @@ public class BeneficiaryApplicationFragment extends BaseFragment implements
      */
     @Override
     public void showBeneficiaryUpdatedSuccessfully() {
-        Toaster.show(rootView, getString(R.string.beneficiary_updated_successfully));
+        String className = BeneficiaryListFragment.class.getName();
+        Fragment fragment = getActivity().getSupportFragmentManager().findFragmentByTag(className);
+        BeneficiaryListFragment castedListFragment = (BeneficiaryListFragment) fragment;
+        castedListFragment.showSnackbar(getString(R.string.beneficiary_updated_successfully));
         getActivity().getSupportFragmentManager().popBackStack();
         getActivity().getSupportFragmentManager().popBackStack();
-
     }
 
 
--- a/app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryDetailFragment.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryDetailFragment.java
@@ -3,6 +3,7 @@ package org.mifos.mobilebanking.ui.fragments;
 import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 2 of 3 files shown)


miPlodder:
@ethan627hsu This is not a good solution. Use coordinator layout, follow this link, 

https://stackoverflow.com/a/32777587

etlhsu:
@miPlodder Coordinador layout doesn't help the situation, the problem is that the snackbar is not bound to the correct layout

etlhsu:
Also, the Coordinator layout implementation would be redundant, as the FAB automatically does the detection of the snackbar

miPlodder:
@ethan627hsu Try that link properly. 

The solution you proposed is not the correct to do these kinds of stuff.

etlhsu:
@miPlodder But how could I implement coordinator layout I'm the list page if the fab is I'm the list page but the snackbar is in the details page? The snackbar would have no reference to the coordiantor layout.

etlhsu:
@miPlodder Do you want me to just change implementation of the snackbar in the list page to include CoordiantorLayout instead of just having it appear?

app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryApplicationFragment.java
```diff
@@ -286,10 +287,12 @@ public void showBeneficiaryCreatedSuccessfully() {
      */
     @Override
     public void showBeneficiaryUpdatedSuccessfully() {
-        Toaster.show(rootView, getString(R.string.beneficiary_updated_successfully));
+        String className = BeneficiaryListFragment.class.getName();
+        Fragment fragment = getActivity().getSupportFragmentManager().findFragmentByTag(className);
+        BeneficiaryListFragment castedListFragment = (BeneficiaryListFragment) fragment;
+        castedListFragment.showSnackbar(getString(R.string.beneficiary_updated_successfully));
```
miPlodder, line 293: Don't create variables, do this without any declaring variable in single line. 
For example `(new Object()).showSnackbar()`.

app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryDetailFragment.java
```diff
@@ -150,7 +151,10 @@ public void onClick(DialogInterface dialog, int which) {
      */
     @Override
     public void showBeneficiaryDeletedSuccessfully() {
-        Toaster.show(rootView, getString(R.string.beneficiary_deleted_successfully));
+        String className = BeneficiaryListFragment.class.getName();
+        Fragment fragment = getActivity().getSupportFragmentManager().findFragmentByTag(className);
+        BeneficiaryListFragment castedListFragment = (BeneficiaryListFragment) fragment;
+        castedListFragment.showSnackbar(getString(R.string.beneficiary_deleted_successfully));
```
miPlodder, line 157: Do the same thing here.

app/src/main/java/org/mifos/mobilebanking/ui/fragments/BeneficiaryListFragment.java
```diff
@@ -65,7 +66,7 @@
     private View rootView;
     private List<Beneficiary> beneficiaryList;
     private SweetUIErrorHandler sweetUIErrorHandler;
-
+    private String snackbarContent = null;
```
miPlodder, line 69: What are these changes in this file ? @ethan627hsu

miPlodder (changes requested):

miPlodder:
@ethan627hsu I just reviewed your PR properly, it appropriate way to do the same. I have requested you some changes. Make those changes.

therajanmaurya:
Hi @ethan627hsu Sorry for the late reply. The solution you people are discussing is totally wrong. When Beneficiary created successfully then show the Toast, not snack bar. Both things have his won significance. 

Make the changes and let me know.

therajanmaurya:
Closing it because of inactivity

PR #946, opened 2018-11-22
Feat: #944 Add a flag to not translatable strings

### LINKED ISSUE #944: Feature Request: Add translatable=false flag to make translations easier to do
**Summary:**

There are some strings which shouldn't be translated.

**Steps to reproduce:**

app>src>main>res>values>strings.xml

**Expected behavior:**

Add translatable="false" flag to non translatable strings.

**Observed behavior:**

There are some strings which shouldn't be translated, but these don't have the flag.

letelete:
Fixes #944 

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1,5 +1,5 @@
 <resources>
-    <string name="app_name">Mifos Mobile</string>
+    <string name="app_name" translatable="false">Mifos Mobile</string>
     <string name="login">Login</string>
     <string name="toast_welcome">Welcome %1$s</string>
     <string name="unable_to_connect">Not connected to internet</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -412,26 +412,26 @@
...
     <!--Format Strings-->
     <string name="hello_client">Hello, %1$s</string>
-    <string name="double_amount">%1$.2f</string>
-    <string name="double_and_string">%1$.2f %2$s</string>
-    <string name="string_and_string">%1$s %2$s</string>
-    <string name="string_and_double">%1$s %2$.2f</string>
-    <string name="string_and_int">%1$s: %2$d</string>
+    <string name="double_amount" translatable="false">%1$.2f</string>
+    <string name="double_and_string" translatable="false">%1$.2f %2$s</string>
+    <string name="string_and_string" translatable="false">%1$s %2$s</string>
```
m-sameer, line 417: For the right to left languages, the order of the two parameters has to be flipped. On the right side will be the first parameter and on the left side will be the second parameter. So I think you should keep it translatable.

app/src/main/res/values/strings.xml
```diff
@@ -412,26 +412,26 @@
...
+    <string name="string_and_double" translatable="false">%1$s %2$.2f</string>
+    <string name="string_and_int" translatable="false">%1$s: %2$d</string>
     <string name="invalid_amount">Invalid Amount</string>
     <string name="exit_message">Press back again to exit</string>
 
     <string name="package_name" translatable="false">package</string>
     <string name="permission_denied_camera">Permission Denied to use Camera</string>
-    <string name="one">1</string>
```
m-sameer, line 425: For some language different numerals are used for example in Hindi, Urdu, Arabic, etc. So this must also be kept as translatable.

letelete:
I'm closing this pr anyway since almost all translations are done.

m-sameer:
Also, I have found a string `pref_configuration` which is used as a key in preferences, therefore it should not be translated to any other language.

m-sameer:
I don't think you should close this PR, we only have translations in about 15 languages. Many of them can be added in future.

m-sameer:
@letelete If you're not working on the issue may I work on it?

letelete:
Sure

PR #951, opened 2018-11-22
Fix #950: Added Portuguese translations

### LINKED ISSUE #950: Feat: Add Portuguese translations
Portuguese translations are needed.

migueldinisdev:
updated /values-pt/strings.xml

Fixes #950 

image

image

### DIFF
--- a/app/src/main/res/values-pt/strings.xml
+++ b/app/src/main/res/values-pt/strings.xml
@@ -1,272 +1,543 @@
-<?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="app_name">Self Service App</string>
-    <string name="login">Login</string>
-    <string name="toast_welcome">Welcome %1$s</string>
-    <string name="unable_to_connect">Not connected to internet</string>
-    <string name="basic">Basic</string>
-    <string name="username">Username</string>
-    <string name="progress_message_login">Logging In</string>
-    <string name="progress_message_loading">Loading</string>
-    <string name="password">Password</string>
+    <string name="app_name">Mifos Mobile</string>
+    <string name="login">Entrar</string>
+    <string name="toast_welcome">Bem-vindo %1$s</string>
+    <string name="unable_to_connect">Não há ligação à internet</string>
+    <string name="basic">Básico</string>
+    <string name="username">Nome de Utilizador</string>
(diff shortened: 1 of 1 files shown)


luckyman20 (commented):
@MigDinny You haven't translated many strings, any specific reason for that?

Also, you haven't added all the strings, compare this file with English one and you will find out what are you missing.

migueldinisdev:
@luckyman20 I didn't notice the pt strings.xml was outdated...

luckyman20:
@MigDinny Please fix the travis build.

app/src/main/res/values-pt/strings.xml
```diff
@@ -1,272 +1,543 @@
...
+    <string name="saving_account_details">Detalhes da Conta Poupança</string>
+    <string name="savings_account">Conta Poupança</string>
+    <string name="error_loan_account_details_loading">Erro ao carregar detalhes da Conta de Empréstimo</string>
+    <string name="nominal_interest_rate">Taxa de Interesse Nominal</string>
+    <string name="account_number">Conta número </string>
+    <string name="account_balance">Saldo da Conta</string>
+    <string name="total_deposits">Depósito Total</string>
+    <string name="open_drawer">Abrir Caixa</string>
+    <string name="close_drawer">Fechar Caixa</string>
     <string name="home">Home</string>
```
luckyman20, line 28: Why you haven't translated this string and many others like this too, any specific reason for that?

migueldinisdev: Yes. They do not have a literal translation and therefore cannot be translated. Example: "Home" translated into portuguese means only "own house" and there is no other word to translate with the same meaning. Besides, "home" is kind of a world loan from English. Most of Portuguese people understand what it is and what it means. We use a lot of English words in our quotidian.

"Splash" for example would be suicide to translate. It can be translated into its literal but nobody would associate it with the real meaning. Most certainly everyone would associate it with this:

image

So, the best option is to keep it untranslated. Like I have said before, in Portugal we use a lot of English words in our everyday. There is no problem.

luckyman20 (approved):
Good Work!

miPlodder:
@MigDinny Good job 👍

PR #966, opened 2018-12-11
Fix #958: Transfer button size shrinks on Confirmation Page

### LINKED ISSUE #958: Transfer button size shrinks on Confirmation Page
**Summary:**

Transfer button size shrinks on Confirmation Page

**Steps to reproduce:**

Complete a Transfer (normal Transfer or 3rd Party Transfer) 

**Expected behaviour:**

Submit button should be properly visible and after transfer complete the cardview is not appropriate.

**Observed behaviour:**

Submit button is not properly visible

Xiaomi, Marshmallow

**Screenshots:**

screenshot_2018-12-03-11-17-42-465_org mifos mobilebanking

etlhsu:
Fixes #958

The problem is caused by the UI not allowing enough space for the close button to appear, it is pushed away by the check icon that appears. I solved the problem by restoring the original scroll state to the app.


miPlodder:
It's not a good design. Instead make https://github.com/openMF/mifos-mobile/blob/0d70dff83db7ef8032067f1823c7e024ed658221/app/src/main/res/layout/fragment_transfer_process.xml#L12 this to match_parent. If it does not work try to add a nestedscrollview.

miPlodder:
Use linearlayout instead of relativelayout.

etlhsu:
@miPlodder You cannot use linearlayout, I could not center the text otherwise

PR #986, opened 2019-01-19
fix: account list is null if no filter is selected

### LINKED ISSUE #985: Account list is null if no filter is selected
**Summary:**
When the user opens filter options, and if he/she doesn't select any options and clicks filter, then the list is null 
**Steps to reproduce:**
1. Open accounts
2. Open filters, don't select any option and click filter
3. We see that the list disappears.

**Expected behaviour:**
If no filter is selected, all the accounts should be displayed.

**Observed behaviour:**

Blank page

shaliniv16:
Fixes #985


shaliniv16:
@Nayan please review.

luckyman20:
@shaliniv16 Please close this, you were neither given a green flag by issue creator nor by any maintainer of the repository. From next time, please follow these guidelines.

We do value your contribution and do appreciate it too but following these guidelines is important for a better systematic process.

PR #996, opened 2019-01-21
Prevented loading of fragment again if its already the current fragment

### LINKED ISSUE #990: Wasteful loading of fragment if its already the selected one
**Summary:**

If you are in a particular fragment (section) then opening navigation drawer and going to same section's menu should only close the navigation drawer.

**Steps to reproduce:**

Open any section like Accounts, Charges etc. Open navigation drawer and click on the selected menu (which should be for the current section). You will see that the app recreates the fragment.

**Expected behaviour:**

The navigation drawer should simply close as the required fragment is already loaded.

**Observed behaviour:**

The app will recreate the fragment and fetch the data again increasing users data usage and also unnecessary rendering. 

Moto G5 Plus, Android 8.1

rahul-jha98:
Fixes #990

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/activities/HomeActivity.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/activities/HomeActivity.java
@@ -85,7 +85,7 @@ public class HomeActivity extends BaseActivity implements UserDetailsView, Navig
     private static final int PLAY_SERVICES_RESOLUTION_REQUEST = 9000;
     private boolean isReceiverRegistered;
     private int menuItem;
-    boolean  doubleBackToExitPressedOnce = false;
+    boolean doubleBackToExitPressedOnce = false;
 
     public void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mifos/mobilebanking/ui/activities/HomeActivity.java
```diff
@@ -154,56 +154,66 @@ protected void onResume() {
...
-                Intent i = new Intent(Intent.ACTION_SEND);
-                i.setType("text/plain");
-                i.putExtra(Intent.EXTRA_TEXT, getString(R.string.string_and_string,
-                        getString(R.string.share_msg), getApplication().getPackageName()));
-                startActivity(Intent.createChooser(i, getString(R.string.choose)));
-                break;
-            case R.id.item_logout:
-                showLogoutDialog();
-                break;
+        //if the selected menu item is checked it means it is the current Fragment
```
luckyman20:
```suggestion
        // If the selected menu item is checked it means it is the current Fragment
```

PR #1003, opened 2019-01-22
fix #1000 Displaying currency code instead of null in Charges fragment

### LINKED ISSUE #1000: Charges and Recent Transaction section shows 'null' in TextView.
**Summary:**

In the charges section, if currency symbol is not present user will see null in UI.
Like "null 50" will be shown as Due amount. 

**Expected behaviour:**

If the symbol is available for the currency then show the symbol. If it is not available, show the currency code like INR, AFN etc. 
I think," **INR 50**" is better than "**null 50**" if rupee symbol is not present.

**Observed behaviour:**

We see "**null 50**" if no currency symbol is present.

Moto G5 Plus, Android 8.1

**Screenshots:**

screenshot_20190122-080824

rahul-jha98:
Fixes #1000 

screenshot_20190122-165916


therajanmaurya:
Good Job. It make sense.

PR #1016, opened 2019-01-23
fix #1009: New instance of home created only if language is changed

### LINKED ISSUE #1009: Back from Settings Activity creates new HomeActivity
**Summary:**

When we go to Settings menu and click back the app creates a new activity. Exiting from this instance of HomeActivity will take you to previous one
The issue was raised before  #891. But the bug is still there.

**Steps to reproduce:**

Go to Settings Menu. Go back to return to HomeActivity. Now press back 2 more times. You will see that the current HomeActitviy closes and a new one opens.

**Expected behaviour:**

It should return to the already created HomeActivity.

**Observed behaviour:**

A new HomeActivity is created on clicking on back button.

Moto G5 Plus, Android 8.1

**Screenshots:**

ezgif-4-fe997b4891c7

rahul-jha98:
Fixes #1009

### DIFF
--- a/app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
@@ -2,12 +2,16 @@ package org.mifos.mobilebanking.ui.activities;
 
 import android.content.Intent;
 import android.os.Bundle;
+import android.support.v4.app.ActivityCompat;
 
 import org.mifos.mobilebanking.R;
 import org.mifos.mobilebanking.ui.activities.base.BaseActivity;
 import org.mifos.mobilebanking.ui.fragments.SettingsFragment;
+import org.mifos.mobilebanking.utils.Constants;
 
 public class SettingsActivity extends BaseActivity {
+
+    private boolean hasSettingsChanged;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -16,12 +20,19 @@ public class SettingsActivity extends BaseActivity {
         setToolbarTitle(getString(R.string.settings));
         showBackButton();
         replaceFragment(SettingsFragment.newInstance(), false, R.id.container);
+        if (getIntent().hasExtra(Constants.HAS_SETTINGS_CHANGED)) {
+            hasSettingsChanged = getIntent().getBooleanExtra(Constants.HAS_SETTINGS_CHANGED,
+                    false);
+        }
     }
 
     @Override
     public void onBackPressed() {
         super.onBackPressed();
-        Intent i = new Intent(this, HomeActivity.class);
-        startActivity(i);
+        if (hasSettingsChanged) {
+            ActivityCompat.finishAffinity(this);
+            Intent i = new Intent(this, HomeActivity.class);
+            startActivity(i);
+        }
     }
 }
--- a/app/src/main/java/org/mifos/mobilebanking/ui/fragments/SettingsFragment.java
+++ b/app/src/main/java/org/mifos/mobilebanking/ui/fragments/SettingsFragment.java
@@ -11,6 +11,7 @@ import android.support.v7.preference.PreferenceFragmentCompat;
 import org.mifos.mobilebanking.R;
 import org.mifos.mobilebanking.utils.ConfigurationPreference;
 import org.mifos.mobilebanking.utils.ConfigurationDialogFragmentCompat;
+import org.mifos.mobilebanking.utils.Constants;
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
```diff
@@ -16,12 +18,24 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        }
     }
 
     @Override
     public void onBackPressed() {
         super.onBackPressed();
-        Intent i = new Intent(this, HomeActivity.class);
-        startActivity(i);
+        if (hasSettingsChanged) {
+            // If Settings have changed, the previous home activity may not be valid
```
miPlodder: Remove the comments.

app/src/main/java/org/mifos/mobilebanking/ui/fragments/SettingsFragment.java
```diff
@@ -68,7 +68,9 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
         if (preference instanceof ListPreference) {
             ListPreference listPreference = (ListPreference) preference;
             LanguageHelper.setLocale(getContext(), listPreference.getValue());
-            startActivity(new Intent(getActivity(), getActivity().getClass()));
+            Intent intent = new Intent(getActivity(), getActivity().getClass());
+            intent.putExtra("hasSettingsChanged", true);
```
miPlodder: Don't use hard-code strings.

app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
```diff
@@ -16,12 +18,24 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            // For example if user changes the language we need to
+            // relaunch previous activity again
+
+            ActivityCompat.finishAffinity(this);
```
miPlodder, line 33: I guess this piece of code is not needed as on pressing the Back button the Activity ends.

app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
```diff
@@ -16,12 +18,24 @@ protected void onCreate(Bundle savedInstanceState) {
         setToolbarTitle(getString(R.string.settings));
         showBackButton();
         replaceFragment(SettingsFragment.newInstance(), false, R.id.container);
+        Intent intent = getIntent();
+        if (intent.hasExtra("hasSettingsChanged")) {
```
miPlodder: Don't use hard-coded strings. Instead, define constants in `Constants.java` and Don't create an object for Intent. Just use `getIntent()` and make the method calls.

app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
```diff
@@ -2,12 +2,14 @@
...
 import android.content.Intent;
 import android.os.Bundle;
+import android.support.v4.app.ActivityCompat;
 
 import org.mifos.mobilebanking.R;
 import org.mifos.mobilebanking.ui.activities.base.BaseActivity;
 import org.mifos.mobilebanking.ui.fragments.SettingsFragment;
 
 public class SettingsActivity extends BaseActivity {
+    private boolean hasSettingsChanged;
```
miPlodder, line 14: Add an empty line before this line.

miPlodder (changes requested):

app/src/main/java/org/mifos/mobilebanking/ui/activities/SettingsActivity.java
rahul-jha98: Yes, but I need to call finishAffinity() for it to work the way I expected. 

Basically, what I have done is that if the user changes the preference for language the hasSettingsChanged boolean becomes true.
If it is true, then after the onbackPressed() call I need to  finish all activities i.e. Settings as well as Home Activity and then launch a new instance of HomeActivity. This new instance will be in the user selected language.

miPlodder:
@rahul-jha98 Add GIF for this (every UI changes). Use DU recorder in your device for creating one. Rest looks good. 👍

rahul-jha98:
When the user **changes the language settings** a new instance of Home Activity is created =>
changed

When the user **simply presses back with no change in language** the Settings Activity simply closes with no wasteful Activity creation.
nochange

miPlodder:
@rahul-jha98  Good job 👍

PR #1114, opened 2019-03-16
fix: add limits in account and phone no in signup

### LINKED ISSUE #1107: Limits for Phone No, Account number in SignUp page
**Summary:**

As the phone numbers are constrained to be 10 digits, and our Account number is also of fixed digits, we should add a check for that before the Sign Up.

**Steps to reproduce:**

Sign Up process

**Expected behaviour:**

User should not be able to Sign Up if the values dont match the size limit.

**Observed behaviour:**

User can Log In at any value.

shiv07tiwari:
Fixes #1107 

Added limits in account number and phone number


shiv07tiwari:
@therajanmaurya @miPlodder Fixed the merge conflict.

therajanmaurya:
No, I don't agree with the fact that phone number hast to 10 digits, we are not building the product for India. it is a global product. and phone number vary according to country.

So we are going to use a country picker library and use it.

PR #1147, opened 2019-03-23
fix: full account no in transfer details

### LINKED ISSUE #1134: Issues in Transaction receipt and preview
**Summary:**

Their are some issues in transaction receipt and transaction review - 
1. No currency with amount
2. Full account numbers are not shown. (1 is shown for 0000000001)

**Steps to reproduce:**

Make a transaction
Screenshot_2019-03-20-23-22-22-219_org mifos mobilebanking

shiv07tiwari:
Fixes #1134 

IMG_20190324_012903

IMG_20190324_012915

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/TransferProcessFragment.java
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/TransferProcessFragment.java
@@ -68,6 +68,7 @@ public class TransferProcessFragment extends BaseFragment implements TransferPro
     private View rootView;
     private TransferPayload payload;
     private TransferType transferType;
+    private String fromAccountNumber, toAccountNumber;
 
     /**
      * Used for TPT Transfer and own Account Transfer.<br>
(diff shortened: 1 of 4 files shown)


shiv07tiwari:
@miPlodder @therajanmaurya Kindly review.
@miPlodder , what I initially did was - 
- Made extra variables  for account numbers in Transfer Payload
- Used those to set the TextViews

But the transfer was giving an error saying another variable in Transfer Payload is not supported.
So I went for this approach.
Is the error because of backend (maybe it has a fixed data about Transfer Payment), or am I missing something ?

shiv07tiwari:
@therajanmaurya @miPlodder Fixed the merge conflict.

app/src/main/java/org/mifos/mobile/ui/fragments/TransferProcessFragment.java
```diff
@@ -77,11 +78,15 @@
      * @param type    enum of {@link TransferType}
      * @return Instance of {@link TransferProcessFragment}
      */
-    public static TransferProcessFragment newInstance(TransferPayload payload, TransferType type) {
+    public static TransferProcessFragment newInstance(
```
therajanmaurya, line 81: Passing two more parameter does not sound good. I would suggest adding two more fields in `TransferPayload`
1. `fromAccountNumber` and 2. `toAccountNumber`

and annotate them with ` @Transient` so it will be ignored in the retrofit call.

therajanmaurya (changes requested):

prashantkh19:
#1176 fixes this. @therajanmaurya close this.

PR #1166, opened 2019-04-10
Fixes 1165: Added support for Two Factor Authentication

### LINKED ISSUE #1165: Support for Two Factor Authentication via Google Authenticator
The app should support two-step verification (time-based preferably) at the time of login. This can be done using Google Authenticator.

prashantkh19:
Fixes #1165
I have made changes to the client side to add this feature. However, API needs to change as we have to store whether the account has TOTP enabled or not, to the server.

Screenshots:
device-2019-04-11-010042
device-2019-04-11-010217

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -67,6 +67,17 @@ android {
     androidExtensions {
         experimental = true
     }
+
+    packagingOptions {
+        exclude 'META-INF/DEPENDENCIES'
+        exclude 'META-INF/LICENSE'
+        exclude 'META-INF/LICENSE.txt'
+        exclude 'META-INF/license.txt'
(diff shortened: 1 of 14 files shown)


shiv07tiwari (commented):
This is already in the GSOC idea list ! This will be done in that period only.

prashantkh19:
> This is already in the GSOC idea list ! This will be done in that period only.

I know that. Actually, I had written some code for my understanding, so I thought I should share it so that I can get some reviews about the approach and it might help someone.

therajanmaurya:
Not implemented properly. as @prashantkh19 is doing GSoC we will pick later and work on it properly.

PR #1175, opened 2019-06-02
Fix #960 : Merge fix of PR 962

### LINKED ISSUE #960: Remove non-translatable strings from translated languages' strings.xml
**Summary:**
Two non-translatable string arrays named **`languages`** and **`languages_value`** are present in the **strings.xml** file for Portuguese.

**Expected behaviour:**
The two arrays should only be present in the default locale and therefore should be removed from **strings.xml** for Portuguese.

prashantkh19:
Fixes #960


prashantkh19:
@therajanmaurya merge this

PR #1176, opened 2019-06-02
Fix #1134 : Requested Changes in PR 1147

### LINKED ISSUE #1134: Issues in Transaction receipt and preview

prashantkh19:
Fixes #1134


therajanmaurya:
@prashantkh19 You are not setting the value in payload, It will always null.

prashantkh19:
> @prashantkh19 You are not setting the value in payload, It will always null.

Fixed

Issue #1096, opened 2019-03-03
Withdrawal and Deposit filter should be added in savings account transaction


theabhishekavi:
**Summary:**

There should be a filter based on withdrawal and deposit

**Steps to reproduce:**

 Open view transactions in saving account account section

**Expected behaviour:**

Based on months there should be one more filter of withdrawal,deposit

prashantkh19:
I would like to work on this.

prashantkh19:
I would like to work on it. I am thinking of making the filter similar to the Account filter.

luckyman20:
Solved in #1184

PR #1184, opened 2019-07-19
Updated filter in SavingAccountsTransactionFragment

### LINKED ISSUE #735: feature:add filters for deposit and withdrawal  in savings account transaction section
**Summary:**
In savings account transactions the list is too long for some account.So it would be better if we add a filter for withdrawal and deposit, it would be much easier go through the transactions.

**Screenshots**
screenshot_20180314-232927

prashantkh19:
Fixes #735 #1096 
### **Screenshots**
image
image


zoman-anamika (approved):
Looks good to me. Good work :)

geekanamika (commented):
As a suggestion, You can Open filter dialog as full-screen dialog instead of pop-up dialog. :)

prashantkh19:
> As a suggestion, You can Open filter dialog as full-screen dialog instead of pop-up dialog. :)

Yeah, but that won't be consistent with the other filter dialogs.

geekanamika:
> > As a suggestion, You can Open filter dialog as full-screen dialog instead of pop-up dialog. :)
> 
> Yeah, but that won't be consistent with the other filter dialogs.

Agreed. We can work on it if we'll think about UI revamp. Replacing pop-up dialogs with Bottom sheet or full screen dialogs depending upon use-case :)

PR #1296, opened 2020-01-25
fix#847 - HomeFragment data shown via shared preferences

### LINKED ISSUE #847: Username and Image should be shown when Internet is not there in Home Fragment

devansh-299:
Fixes #847 

Screen recording

GIF-200125_200901

### DIFF
--- a/app/src/main/java/org/mifos/mobile/api/local/PreferencesHelper.java
+++ b/app/src/main/java/org/mifos/mobile/api/local/PreferencesHelper.java
@@ -33,6 +33,8 @@ public class PreferencesHelper {
     private static final String TENANT = "preferences_base_tenant";
     private static final String BASE_URL = "preferences_base_url_key";
     private static final String PROFILE_IMAGE = "preferences_profile_image";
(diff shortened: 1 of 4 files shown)


luckyman20:
Good work here and it is unfortunate that we could not get it reviewed any sooner. @gururani-abhishek will be taking up this issue and do the same changes for the respective kotlin files.

PR #1323, opened 2020-02-07
Fixes #1322 - removed manual login dialog

### LINKED ISSUE #1322: Cancel manual login in PassCodeActivity does not work
**Summary:**

When the manual login dialog is cancelled in the `PassCodeActivity`, even though the user is not redirected to login manually the app does not recognize the user's passcode. This happens because in `MifosPasscodeActivity` `clearTokenPreferences` is called before `startLoginActivity`. I think we should get rid of the login dialog altogether because making the cancel dialog work would entail modifying `MifosPasscodeActivity`.

**Steps to reproduce:**

In the `PassCodeActivity` press 'forgot passcode, login manually' then cancel the dialog then try logging in with your passcode. 

**Screen Recording**

dubeyaayush07:
Fixes #1322
Removed manual login dialog, now pressing forgot passcode directly launches the LoginActivity

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/activities/PassCodeActivity.java
+++ b/app/src/main/java/org/mifos/mobile/ui/activities/PassCodeActivity.java
@@ -1,7 +1,6 @@
 package org.mifos.mobile.ui.activities;
 
 import android.Manifest;
-import android.content.DialogInterface;
 import android.content.Intent;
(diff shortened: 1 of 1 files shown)


PratyushSingh07:
we can close this PR @therajanmaurya . This issue no longer exits in the app

PR #1340, opened 2020-02-13
Fix #1337 -  Loan transactions RecyclerView height and margins refactored

### LINKED ISSUE #1337: Loan transactions Recycler view
**Summary:**

The loan transactions recyclerview does not cover the entire screen which ideally it is supposed to do . Currently it covers only the area required area to contain the items inside it. Hence swiping across them results in bad user experience

**Steps to reproduce:**
* Go to accounts -> Loan tab
* Select any disbursed loan
* Tap on Transactions

**Expected behaviour:**

Ideally the recycler view should have covered the entire remaining portion of the screen

Xiaomi PocoF1
Android 9
Stock Rom

jawidMuhammadi:
Fixes #1337 

Loan transactions RecyclerView height and margins refactored. Because it was not covering the entire screen which ideally it is supposed to do. Swiping across them results in bad user experience and also it was not according to material design guidelines.
**Before:**
Screenshot_2020-02-13-23-57-06
Screenshot_2020-02-13-23-57-13

**Now:**
Screenshot_2020-02-13-23-41-59
Screenshot_2020-02-13-23-45-04
#1337

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ ext {
     buildToolsVersion = '28.0.3'
 
     // App dependencies
-    supportLibraryVersion = '1.1.0'
+    supportLibraryVersion = '1.0.0'
     designLibraryVersion = '1.0.0'
     daggerVersion = '2.5'
     retrofitVersion = '2.2.0'
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ ext {
     buildToolsVersion = '28.0.3'
 
     // App dependencies
-    supportLibraryVersion = '1.0.0'
+    supportLibraryVersion = '1.1.0'
     designLibraryVersion = '1.0.0'
     daggerVersion = '2.5'
     retrofitVersion = '2.2.0'
--- a/app/src/main/res/layout/fragment_loan_account_transactions.xml
+++ b/app/src/main/res/layout/fragment_loan_account_transactions.xml
@@ -13,8 +13,7 @@
 
     <LinearLayout
         android:id="@+id/ll_loan_account_trans"
-        android:layout_height="wrap_content"
-        android:layout_margin="@dimen/default_margin"
+        android:layout_height="match_parent"
         android:layout_width="match_parent"
         android:orientation="vertical"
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.java
```diff
@@ -547,6 +547,8 @@ public void onItemSelected(AdapterView<?> parent, View view, int position, long
 
             case R.id.sp_loan_purpose:
                 if (loanTemplate.getLoanPurposeOptions() != null) {
+                    if (position == loanTemplate.getLoanPurposeOptions().size())
+                        position--;
```
naman14: can you explain why are we doing this?

build.gradle
```diff
@@ -38,7 +38,7 @@ ext {
     buildToolsVersion = '28.0.3'
 
     // App dependencies
-    supportLibraryVersion = '1.1.0'
+    supportLibraryVersion = '1.0.0'
```
naman14: remove this change, keep separate PRs for each issue/change

app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.java
jawidMuhammadi: Actually this line of code was regarding to another issue that has been raised by me. During committing I have written the wrong issue number in the commit message. I did update the commit message but I don't know why the change didn't appear here.

build.gradle
jawidMuhammadi: Ok I will. But if I revert this change, the recycler items in saving account, loan account and share account do not work. Before pushing my code I asked the community one of them advised to not remove this change.

app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.java
jawidMuhammadi: I removed this change. I find out it has already been done by another developer.

PR #1350, opened 2020-02-14
Fix #1349 - Removed error indicator when valid value is entered

### LINKED ISSUE #1349: Error indicator in LoanApplicationFragment is not removed even after the loan is reviewed
**Summary:**
The error indicator in `tilPrincipalAmount` persists even after correct value is entered and is visible when coming back from the review window.

**Steps to reproduce:**
Enter an invalid value now correct the value and press the review button, now press back the error is still there

**Screen Recording**

dubeyaayush07:
Fixes #1349 


naman14:
looks good

PR #1359, opened 2020-02-15
Displaying currency code when currency symbol is null and displaying 0.00 when principal value is null in Repayment Schedule TableView

### LINKED ISSUE #1355: Displaying null in Repayment Schedule Table View
**Summary:**

Displaying null value instead of currency symbol or currency code in Repayment Schedule Fragment's table view and also it is showing nu when the amount is 0 or null in Loan Balance column in the table view

**Steps to reproduce:**

- Go to accounts -> Loan tab
- Select any disbursed loan
- Tap on Repayment Schedule

**Expected behaviour:**

The app should show a currency symbol or a currency code in case if there is no symbol for it -like AFN beside amount value and if the amount value is 0 or null it should show as 0.00, not _nu_

Pixel 3, API 29
Android Version 10 (Q)
Emulator

**Screenshots:**

Screenshot_1581763081

jawidMuhammadi:
Fixes #1355 

Screenshot_1581800485


naman14:
Looks good

PR #1393, opened 2020-03-08
Fix-#1392 Updated Settings Toolbar Title on Back Button Pressed

### LINKED ISSUE #1392: Settings ToolbarTitle not changed on BackButton Pressed
**Summary:**

On the back button pressed from the change password, the toolbar title is not changed.

**Expected behaviour:**

The toolbar title of settings activity should change on the back button pressed from the change password.

**Observed behaviour:**

The toolbar is not changed and is the same on the back button pressed.

Vivo V17
Android 9.0

**ScreenShot:**

ezgif-3-7b07066b88ff

rutvik-panchal:
Fixes #1392 
Updated Toolbar Title on the back button pressed from the change password window.

**Screenshots:**

ezgif-3-42bb9c5a8e1d
Updated Toolbar Title on the back button pressed from the change password window.
ing them.


app/src/main/java/org/mifos/mobile/ui/fragments/SettingsFragment.java
```diff
@@ -36,6 +36,7 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
     @Override
     public void onResume() {
         super.onResume();
+        getActivity().setTitle("Settings");
```
jawidMuhammadi: Please remove hardcoded "Settings" strings from here. Using hardcoded strings is not a good practice. First, add the string in the `strigs.xml` file then use it dynamically.

jawidMuhammadi (changes requested):

app/src/main/java/org/mifos/mobile/ui/fragments/SettingsFragment.java
rutvik-panchal: Yes, I Will take care of that in the future & will update it soon.

PR #1413, opened 2020-03-14
Fix error #1408

### LINKED ISSUE #1408: Error in EditText View on Login Page doesn't get removed on entering data
Summary : If user clicks on login button without entering the username and password i.e leaving both fields empty then error is shown. When user clicks on the username field to enter valid data the error doesn't get removed. Same is the case with password field.

Steps to reproduce : Listener is to be kept so that clicking on the respective field, the error of the respective field gets removed.

Expected behaviour : Once error is shown, user learns that username and password has to be entered so if he/she clicks on the respective fields the respective errors should be removed.

Observed behaviour : On clicking the respective username and password fields, the errors doesn't get removed.

Device and Android version : One Plus 6 (Android 10)

Screenshots :

virajpatel2411:
Fixes #1408 
Screenshots: 


gradle.properties
```diff
@@ -21,6 +21,7 @@ org.gradle.jvmargs=-XX:MaxPermSize=2048m -XX:+CMSClassUnloadingEnabled -XX:+Heap
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 org.gradle.parallel=true
+org.gradle.java.home=C:\\Program Files\\Java\\jdk1.8.0_161
```
iamsh4shank: Why this change?

iamsh4shank (changes requested):

gradle.properties
virajpatel2411: Else the gradlew check was failing.
I searched the reason for error and did the change.
Should I revert it?

PR #1498, opened 2020-06-08
fix: Smooth movement between fragment transition

### LINKED ISSUE #1496: No smooth transitions for fragments
**Summary:**
The transition between fragments in the application is not smooth and takes time to load. We can make the transition smooth while switching between the previous and next fragments.

Vivo 1724
Android 8.1.0
Funtouch OS_4.0

ShivangiSingh17:
Fixes #1496 

- [X] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [X] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [X] If you have multiple commits please combine them into one commit by squashing them.

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/activities/base/BaseActivity.java
+++ b/app/src/main/java/org/mifos/mobile/ui/activities/base/BaseActivity.java
@@ -26,6 +26,7 @@ import androidx.core.view.ViewCompat;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
 import androidx.fragment.app.FragmentTransaction;
+import androidx.transition.Fade;
 
 /**
  * @author ishan
@@ -69,6 +70,7 @@ public class BaseActivity extends BasePassCodeActivity implements BaseActivityCa
 
     /**
      * Used for dependency injection
+     *
      * @return {@link ActivityComponent} which is used for injection
      */
     public ActivityComponent getActivityComponent() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mifos/mobile/ui/activities/base/BaseActivity.java
```diff
@@ -214,15 +215,29 @@ public void replaceFragment(Fragment fragment, boolean addToBackStack, int conta
...
             if (addToBackStack) {
                 transaction.addToBackStack(backStateName);
             }
-            transaction.commit();
+            transaction.commitAllowingStateLoss();
         }
     }
 
+    public void transition(Fragment previousFragment, Fragment nextFragment){
+        if(previousFragment!=null) {
```
jawidMuhammadi: After `if` and `!=` one white space is needed that is why your build is failing. You can simply solve it by formating the code by pressing `ctrl+alt+l` :)

app/src/main/java/org/mifos/mobile/ui/activities/base/BaseActivity.java
```diff
@@ -214,15 +215,29 @@ public void replaceFragment(Fragment fragment, boolean addToBackStack, int conta
...
+            Fade exitFade = new Fade();
+            exitFade.setDuration(250);
```
jawidMuhammadi: Instead of using hardcoded values, what about if we use standard timing for animations suggested by Googlers like this:
`exitFade.setDuration(android.R.integer.config_shortAnimTime)`

jawidMuhammadi (changes requested):

PR #1526, opened 2020-07-16
Java to Kotlin conversion

ashwinkey04:
Solves Issue(s):  #1530 

Converted the following classes which were previously in Java, to Kotlin.

1. Passcode activity
2. Constants
3. Convert Registration activity
4. Home activity and fragments
5. Base activity and fragments
6. API related classes 
7. All View classes
8. All classes under `/main/../ui` (Activities, Adapters, Enums, Views, Widgets, Fragments)
9. All Adapter classes
10. All Presenter classes
11. All JUnit tests 
12. All activity classes
13. All model classes
14. All Injection classes
15. MifosSelfService class
16. commonTest and androidTest classes 

---

- [ ] If you have multiple commits please combine them into one commit by squashing them 
(Separate commits for easier reviews. Will squash after the final review)


app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
```diff
@@ -0,0 +1,60 @@
...
+    @kotlin.jvm.JvmField
+    @BindView(R.id.tv_copy_right)
+    var tvCopyRight: TextView? = null
+    var rootView: View? = null
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
+                              savedInstanceState: Bundle?): View? {
+        rootView = inflater.inflate(R.layout.fragment_about_us, container, false)
+        ButterKnife.bind(this, rootView!!)
+        setToolbarTitle(getString(R.string.about_us))
+        tvAppVersion!!.text = getString(R.string.app_version, BuildConfig.VERSION_NAME)
```
jawidMuhammadi: I think it is much better to use the safe call `?.` (One of the biggest advantages of using Kotlin) instead of `!!`, because it leads the App to crash if our view or any other item is null or it has not been initiated. You can replace all `!!` to `?.` in a file, simply by entering `Ctlr+r`.

jawidMuhammadi (changes requested):

app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.mifos.mobile.ui.fragments.base.BaseFragment
+import java.util.*
+
+/*
+~This project is licensed under the open source MPL V2.
+~See https://github.com/openMF/self-service-app/blob/master/LICENSE.md
+*/   class AboutUsFragment : BaseFragment() {
+    @kotlin.jvm.JvmField
+    @BindView(R.id.tv_app_version)
+    var tvAppVersion: TextView? = null
```
jawidMuhammadi, line 30: If we are using **Kotlin** then isn't it better to optimize our code a little more and decrease lines of code. Like, here, I think there is no need for those 3 lines of code(23,24,25) and in all other places. We can simply call `tv_app_version` whenever we want using **Kotlin android-synthetics**

jawidMuhammadi (changes requested):

app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
ashwinkey04: Thank you for your insights @jawidMuhammadi. I guess I had failed to notice a lot of not-null assertion operators since this conversion was straight from Java. Though there are quite a few functions in the app which requires a non-null object (For instance, `ButterKnife.bind()` is a Java function and doesn't support null safety yet), I would check for usages where safe call is possible and update it soon. Thanks again :)

app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
ashwinkey04: Sorry for the late reply on this, I just noticed this comment. Firstly, KotlinX android synthetics is now regarded to be not a recommended practice anymore [[1]](https://android-review.googlesource.com/c/platform/frameworks/support/+/882241) [[2]](https://old.reddit.com/r/androiddev/comments/ala9p2/why_kotlinx_synthetic_is_no_longer_a_recommended/efdvpkg/) since it was replaced with `findViewById`. I did not use that either since the app already used ButterKnife for ViewBinding and I had to primarily focus on completely converting the app to Kotlin first. I will discuss and get back on if I should replace ButterKnife. Thank you for the suggestions and do let me know if there are any more optimization suggestions from your side :)

app/src/main/AndroidManifest.xml
```diff
@@ -56,7 +56,8 @@
         <activity android:configChanges="orientation|screenSize"
                   android:launchMode="singleTask"
                   android:name="org.mifos.mobile.ui.activities.PassCodeActivity"
-                  android:windowSoftInputMode="adjustResize"/>
```
miPlodder, line 59: Why is this `screenOrientation` added. Please clarify

ashwinkey04: If the orientation is landscape in the passcode activity, the UI elements go out of bounds. mifos-mobile-cn also had a similar [problem](https://github.com/openMF/mifos-mobile-cn/issues/108) which was solved by restricting the orientation to only portrait.

edcable:
@miPlodder is this ready to be merged?

miPlodder:
@edcable This PR looks good to me.

@therajanmaurya If possible can you also take a look at it once and share your thoughts?

PR #1545, opened 2020-08-27
feat: Add UI Tests for Login flow with Espresso

ashwinkey04:
Uses the espresso library to perform UI tests for complete login flow
<p align="center">
</p>
</br>

Steps implemented:
- Enter username: _ashwinr_
- Enter password: _password_
- Tap password visibility toggle twice
- Press login button
- Wait for login request to complete (Uses `CountingIdlingResource`)
- Checks if passcode activity layout is properly loaded and displayed
- Tap password visibility toggle 
- Perform swipe up action to make the digit _0_ and _proceed_ buttons visible in smaller devices (Required for Travis CI to pass)
- Tap digit _0_ four times (To input passcode as 0000)
- Press proceed button
- Perform passcode input again
- Press save button

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -68,6 +68,9 @@ android {
     androidExtensions {
         experimental = true
     }
+    testOptions {
+        animationsDisabled = true
+    }
 }
 
 dependencies {
@@ -139,6 +142,9 @@ dependencies {
 
     //TableView
     implementation "com.evrencoskun.library:tableview:$rootProject.tableViewVersion"
+
+    //Espresso Idling Resource
+    implementation "androidx.test.espresso:espresso-idling-resource:$rootProject.espressoVersion"
 
     // Unit tests dependencies
     testImplementation "junit:junit:$rootProject.jUnitVersion"
--- a/app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
+++ b/app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
@@ -1,0 +1,192 @@
+package org.mifos.mobile.ui.activities
+
+import android.view.View
+import android.view.ViewGroup
+
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.IdlingRegistry
+import androidx.test.espresso.ViewInteraction
+import androidx.test.espresso.action.ViewActions.*
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.*
+import androidx.test.filters.LargeTest
+import androidx.test.rule.ActivityTestRule
+import androidx.test.rule.GrantPermissionRule
+import androidx.test.runner.AndroidJUnit4
+
+import org.hamcrest.Description
+import org.hamcrest.Matcher
+import org.hamcrest.Matchers.`is`
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.TypeSafeMatcher
+
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mifos.mobile.R
+import org.mifos.mobile.utils.EspressoIdlingResouce
+
+/**
+ * UI Testing for Login using Espresso
+ *
+ * @author Ashwin Ramakrishnan
+ * @since 22/08/2020
+ */
+
+@LargeTest
+@RunWith(AndroidJUnit4::class)
+class LoginSetPasscodeTest {
+
+    @Rule
+    @JvmField
+    var mActivityTestRule = ActivityTestRule(SplashActivity::class.java)
+
+    @Rule
+    @JvmField
+    val grantPermissionRule: GrantPermissionRule =
+        GrantPermissionRule.grant(android.Manifest.permission.READ_PHONE_STATE)
+
+    @Before
+    fun registerIdlingResource() {
+        IdlingRegistry.getInstance().register(EspressoIdlingResouce.countingIdlingResource)
+    }
+
+    @After
+    fun unregisterIdlingResource() {
+        IdlingRegistry.getInstance().unregister(EspressoIdlingResouce.countingIdlingResource)
+    }
+
+    @Test
+    fun loginSetPasscodeTest() {
+        val usernameEditText = onView(
+            allOf(
+                childAtPosition(
+                    childAtPosition(
+                        withId(R.id.til_username),
+                        0
+                    ),
+                    0
+                )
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/mifos/mobile/ui/activities/LoginActivity.kt
```diff
@@ -88,6 +89,7 @@ class LoginActivity : BaseActivity(), LoginView {
      * Starts [PassCodeActivity]
      */
     override fun showPassCodeActivity() {
+        EspressoIdlingResouce.decrement()
```
miPlodder, line 92: I don't think, for writing integration tests or be it unit test we will have to make changes in the testing classes, check on this.

app/build.gradle
```diff
@@ -68,6 +68,9 @@ android {
     androidExtensions {
         experimental = true
     }
+    testOptions {
```
miPlodder, line 71: Did this change fix the Travis build issue, if yes then please share the reference?

app/build.gradle
```diff
@@ -140,6 +143,9 @@ dependencies {
     //TableView
     implementation "com.evrencoskun.library:tableview:$rootProject.tableViewVersion"
 
+    //Espresso Idling Resource
+    implementation "androidx.test.espresso:espresso-idling-resource:$rootProject.espressoVersion"
```
miPlodder, line 147: Was `expressoVersion` already existing?

app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
```diff
@@ -0,0 +1,192 @@
...
+        IdlingRegistry.getInstance().register(EspressoIdlingResouce.countingIdlingResource)
+    }
+
+    @After
+    fun unregisterIdlingResource() {
+        IdlingRegistry.getInstance().unregister(EspressoIdlingResouce.countingIdlingResource)
+    }
+
+    @Test
+    fun loginSetPasscodeTest() {
```
miPlodder, line 62: Is it an integration test for a single flow, clarify. If not split it multiple flows into multiple tests, instead of covering all flows in a single test method.

miPlodder (changes requested):

app/src/main/java/org/mifos/mobile/ui/activities/LoginActivity.kt
ashwinkey04: Using `EspressoIdlingResource` demands production-level code be altered for letting Espresso know that the UI is being updated due to a task. This is preferred over sleeping the thread in the test resource. More information is available in the [documentation](https://developer.android.com/reference/androidx/test/espresso/idling/CountingIdlingResource/)

app/build.gradle
ashwinkey04: It is a recommended practice to disable animations in the device for testing. It is also mentioned [here](https://developer.android.com/training/testing/espresso/setup#set-up-environment) in the  documentation for espresso

app/build.gradle
ashwinkey04: Yes. The test implementation was also previously mentioned [here](https://github.com/openMF/mifos-mobile/blob/95cc2fd10d3a62509d4fb7fd1ffeaae868f6c120/app/build.gradle#L164)

app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
ashwinkey04: This test is meant to cover the login flow completely. So I have included steps to set passcode too so that the complete flow is covered.

app/src/main/java/org/mifos/mobile/ui/activities/LoginActivity.kt
miPlodder: Are we making an actual API call, during the Integration test? If yes, let's not do it.
We can mock the API call, with some hardcoded JSONs, check on the feasibility of the solution which I'm referring to in this comment?

Else we can go with this approach `EspressoIdlingResource` also. Moreover, we need to add `EspressoIdlingResource` at [Service](https://github.com/openMF/mifos-mobile/blob/development/app/src/main/java/org/mifos/mobile/presenters/LoginPresenter.kt#L53) level, just before the call increment it, and inside `success` method or `failure` method decrement it, adding it to UI level will involve lots of cases that needs to be taken care of.

app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
miPlodder: Sounds good.
Let's go with a failure case, where login is a failure. `Invalid creds`

Suggested naming convention, add suffix `_validCreds` and `_invalidCreds`. You can keep the prefix same for both the test methods.

app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
```diff
@@ -0,0 +1,192 @@
...
+                    childAtPosition(
+                        withId(R.id.til_username),
+                        0
+                    ),
+                    0
+                )
+            )
+        )
+
+        val passwordEditText = onView(
```
miPlodder, line 75: Let's follow convention, don't add a new line after every variable creation or after every operation. Let's add lines as a delimiter for better readability, add them wisely.

app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
ashwinkey04: Yes, will make those changes

app/src/androidTest/java/org/mifos/mobile/ui/activities/LoginSetPasscodeTest.kt
ashwinkey04: I'll look into those and remove them

app/src/main/java/org/mifos/mobile/ui/activities/LoginActivity.kt
ashwinkey04: I will check for the feasibility and update on it sooner. Meanwhile, I will mark this PR as a draft

miPlodder:
@ashwinkey04 Do we have any progress on it?

ashwinkey04:
I was able to set up a mock web server using a library from OkHttp. But it was possible to place only a single API call using that. I'll need to figure something out to mock all the APIs used in the app

PR #1579, opened 2020-11-28
Remove unnecessary ndk specifications from build.gradle files

### LINKED ISSUE #1578: Remove unnecessary ndk specifications from build.gradle files
**Summary:**
There is no need to change the build.gradle files for Github Actions CI build. We can safely remove them.

Sparsh1212:
Fixes #1578

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,8 +10,7 @@ apply from: '../config/quality/quality.gradle'
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
     buildToolsVersion rootProject.ext.buildToolsVersion
-    ndkVersion rootProject.ext.ndkVersion
-
+    
     defaultConfig {
         applicationId "org.mifos.mobile"
         minSdkVersion rootProject.ext.minSdkVersion
--- a/build.gradle
+++ b/build.gradle
@@ -36,8 +36,7 @@ ext {
     targetSdkVersion = 28
     compileSdkVersion = 28
     buildToolsVersion = '28.0.3'
-    ndkVersion = '21.3.6528147'
-
+    
     // App dependencies
     supportLibraryVersion = '1.0.0'
     designLibraryVersion = '1.0.0'


garvit984:
@devansh-299 Can you please review it?

app/build.gradle
```diff
@@ -10,8 +10,7 @@ apply from: '../config/quality/quality.gradle'
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
     buildToolsVersion rootProject.ext.buildToolsVersion
-    ndkVersion rootProject.ext.ndkVersion
-
+
```
miPlodder, line 13: This removal makes sense to me. I might have missed it, was in rush yesterday.

@rutvik-panchal 
- Can you clarify what was the reason behind the addition of this (since integrating workflows should be independent of the app)? (were you facing some issues, while testing it in your fork)
- Similarly, what's the reason for adding it to the workflow too? If possible, share some reference link.

miPlodder (commented):
Good catch, This removal makes sense to me. I might have missed it, in the previous PR, was in a rush yesterday.

app/build.gradle
Sparsh1212: @miPlodder
 We only need to specify the ndk version either in the workflow file or the build.gradle files. And since it has already been specified in the workflow, so we don't need to specify it explicitly in the build.gradle files. Here is the reference for it: [Link](https://github.community/t/android-ndk-on-unix-is-incompatible-with-android-build-tools-3-6-0/17697/15)
Actually github actions require ndk version to be specified for the check to run. It is a known issue, not fixing it causes the execution for stripDebugDebugSymbols fail which in turn fails entire process. Official Docs: [here](https://developer.android.com/studio/projects/install-ndk#apply-specific-version).

rutvik-panchal: It's an [issue](https://github.community/t/android-ndk-on-unix-is-incompatible-with-android-build-tools-3-6-0/17697/13) and i have followed a comment ref. [link](https://github.community/t/android-ndk-on-unix-is-incompatible-with-android-build-tools-3-6-0/17697/15), If we are using gradle version < 4.1 then we need to specify the ndkVersion in build.gradle and we are currently at version 3.3.2 ([version code link](https://github.com/openMF/mifos-mobile/blob/d9fd53cd59a79cc33858133effc8e4f393a0bc4e/build.gradle#L10)).
I will check it once and let you know.

rutvik-panchal: We can remove it, I have tested on my forked repository [actions link](https://github.com/rutvik-panchal/mifos-mobile/actions/runs/389808022) for [changes](https://github.com/rutvik-panchal/mifos-mobile/commit/256ac30412ce4f087c101b86468dbe507f1af258).

miPlodder:
Can we follow commit-msg standards properly, I see them being violated here and correct the typo for `gradle`. Refer https://github.com/openMF/mifos-mobile/wiki/Commit-Style-Guide
Rest looks fine to me

Sparsh1212:
> Can we follow commit-msg standards properly, I see them being violated here and correct the typo for `gradle`. Refer https://github.com/openMF/mifos-mobile/wiki/Commit-Style-Guide
> Rest looks fine to me

@miPlodder Done!

Issue #1573, opened 2020-11-26
App crashes on pressing back button on review loan page 


PatelVatsalB21:
**Summary:**
On Review loan page if back button from toolbar or back button is pressed app crashes. Problem is with Purpose of Loan spinner

**Expected behaviour:**
It should load all selected and entered values on Submit loan page

**Device and Android version:**
Lenovo Tab7 
Android 7.0
Stock ROM

![Screenshot_20201126-235513](https://user-images.githubusercontent.com/70195106/100383299-4c8a3480-3043-11eb-8cee-0d81eceb9cfc.jpeg)
![Screenshot_20201126-235526](https://user-images.githubusercontent.com/70195106/100383311-52801580-3043-11eb-8481-36d6fde0874d.jpeg)

PatelVatsalB21:
can i work on this issue?

luckyman20:
Unable to reproduce this issue.

PR #1597, opened 2020-11-30
Fixed crash on Review Loan page #1573

PatelVatsalB21:
Fixes  bug #1573  that made app crash on pressing back button

**Stacktrace**
Capture

**Proof of Fix**

GIF-201206_123850

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.kt
@@ -1,6 +1,7 @@
 package org.mifos.mobile.ui.fragments
 
 import android.os.Bundle
+import android.os.Handler
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -497,7 +498,7 @@ class LoanApplicationFragment : BaseFragment(), LoanApplicationMvpView, OnDatePi
                             LoanState.UPDATE)
                 }
             }
-            R.id.sp_loan_purpose -> purposeId = loanTemplate?.loanPurposeOptions?.get(position)?.id
+            R.id.sp_loan_purpose -> purposeListLoader(position)
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.kt
```diff
@@ -508,6 +509,21 @@ class LoanApplicationFragment : BaseFragment(), LoanApplicationMvpView, OnDatePi
         loanApplicationPresenter?.detachView()
     }
 
+    private fun purposeListLoader(position: Int) {
+        if (loanTemplate?.loanPurposeOptions?.size == 0) {
+            loanApplicationPresenter?.loadLoanApplicationTemplateByProduct(productId,
+                    LoanState.UPDATE)
+            val handler = Handler()
```
miPlodder, line 516: Why are we posting delays and making the recursive call? Can we come up with a better solution or can you explain why you came up with this fix ?

miPlodder (changes requested):
- Can we add stacktrace for crash ?
- Follow standard commits wiki 
- can we also add ss or gif as a proof of concept for the fix

app/src/main/java/org/mifos/mobile/ui/fragments/LoanApplicationFragment.kt
PatelVatsalB21: I wasn't able to find other solution for a long time, but i noticed that loading list of purpose options takes some time so i used handler and recursion is used as based on phone's capability if low memory phone takes longer that specified it will add some more time. If you are not satisfied with results, i will try to change code but need some guidance, because i tried long time and on every method but came no results.

PatelVatsalB21:
ss had no point as it showed only app crash dialog and homescreen,but i will attach it with stacktrace ASAP:)

PatelVatsalB21:
I will attach stacktrace of error and gif of fix ASAP and will also take care of following commit style next time. Sorry:)

PatelVatsalB21:
We have #1681 as better option so i am closing this pr

PR #1642, opened 2020-12-23
Fixes_#1641: Alignment of circular imageView in User Details fixed

### LINKED ISSUE #1641: Bad alignment of text drawable in User Details Section
**Summary:**
Alignment of text drawable circular imager view is bad, it is at bottom which does not matches with background

**Expected behaviour:**
It should be center vertical

Lenovo Tab 7
Android 7

**Screenshots:**
ss

PatelVatsalB21:
Fixes #1641 
Alignment of imageview has be set to `center_vetical` along with "10dp" `margin_top`
ss

### DIFF
--- a/app/src/main/res/layout/fragment_user_profile.xml
+++ b/app/src/main/res/layout/fragment_user_profile.xml
@@ -29,7 +29,7 @@
                 tools:src="@drawable/mifos_logo"/>
 
             <LinearLayout
-                android:gravity="bottom"
+                android:gravity="center_vertical"
                 android:layout_height="match_parent"
                 android:layout_margin="@dimen/margin_32dp"
                 android:layout_width="150dp"


app/src/main/res/layout/fragment_user_profile.xml
```diff
@@ -39,6 +39,7 @@
                     android:id="@+id/iv_text_drawable"
                     android:layout_height="84dp"
                     android:layout_width="84dp"
+                    android:layout_marginTop="10dp"
```
miPlodder: why do we need `marginTop`, won't changing the gravity serve the purpose? clarify

miPlodder (changes requested):
- can we remove the underscore from commit-msg also?

app/src/main/res/layout/fragment_user_profile.xml
PatelVatsalB21: @miPlodder as we can see in this image attached below, linearLayout that is parent of Imageview with `@+id/iv_text_drawable` has height set to `match parent` with `32dp` margin so it intersects with toolbar. so if i put only gravity as `center_vertical` user will see offset and more towards top instead of center as visible in image(smallest square), so i added offset of `marginTop="10dp"`.

UserProfile

miPlodder: if you ought to compare, add ss for both the cases

PatelVatsalB21: @miPlodder can you feel difference?

without marginTop
ss

with marginTop
ss

PatelVatsalB21: @miPlodder are there any changes i need to apply?

miPlodder: remove additional margin, different device sizes will have a different orientation for it. Let's default it `center_vertical` only

PatelVatsalB21: @miPlodder please review changes

PatelVatsalB21:
@garvit984 I have already requested review from @miPlodder.

PR #1718, opened 2021-02-13
Fixes #1717: Beneficiary creation and update fixed

### LINKED ISSUE #1717: Unable to create or update Beneficiary
**Summary:**
As user creates new beneficiary after entering all details, it shows same error, failing to create beneficiary. Same issue happens when user tries to update beneficiary

**Steps to reproduce:**
1 Go to Beneficiary Section
2 Create New Beneficiary

**Expected behaviour:**
Beneficiary should be created successfully

Any

**Create Beneficiary:**
https://user-images.githubusercontent.com/70195106/107857316-ccaa8400-6e53-11eb-9f0b-d0103d2faa4a.mp4

**Update Beneficiary:**
https://user-images.githubusercontent.com/70195106/107857310-c0bec200-6e53-11eb-9ff1-6e6126d9f9c2.mp4

PatelVatsalB21:
Fixes #1717 

https://user-images.githubusercontent.com/70195106/107856569-fb722b80-6e4e-11eb-9c47-0f86b46aadc7.mp4

### DIFF
--- a/app/src/main/java/org/mifos/mobile/models/beneficiary/BeneficiaryPayload.kt
+++ b/app/src/main/java/org/mifos/mobile/models/beneficiary/BeneficiaryPayload.kt
@@ -17,7 +17,7 @@ data class BeneficiaryPayload(
 
         var accountType: Int? = 0,
 
-        var transferLimit: Float? = 0f,
+        var transferLimit: Int? = 0,
 
         var officeName: String? = null
 ) : Parcelable
--- a/app/src/main/java/org/mifos/mobile/models/beneficiary/BeneficiaryUpdatePayload.kt
+++ b/app/src/main/java/org/mifos/mobile/models/beneficiary/BeneficiaryUpdatePayload.kt
@@ -6,5 +6,5 @@ package org.mifos.mobile.models.beneficiary
 
 data class BeneficiaryUpdatePayload @JvmOverloads constructor(
         var name: String? = null,
-        var transferLimit: Float = 0f
+        var transferLimit: Int = 0
 )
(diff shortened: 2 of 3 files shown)


divyank00:
@PatelVatsalB21 As mentioned in one of the previous or, this issue must be handled in the backend so this issue needs to be raised in the backend repo. The amount can be a floating number.

PatelVatsalB21:
@divyank00 yes you are right but here backend issue are long awaited, one for editing profile is there but no endpoint for editing profile is available, so during GSOC period mentor and maintainers would be active we can ask at that time. Or do you say to raise error in fineract ?

afk11:
@divyank00 @PatelVatsalB21 I ran into this issue, and was about to create an issue here. If I understand correctly, the server never accepted `float` before, so the app should send `long` until the server is changed. It's currently not possible to send `1000` or `1000.00` using the app so the feature is totally broken. IMO this PR should be merged, and an issue created on the fineract JIRA to request the new feature if desirable (accepting floats)

Create Beneficiary validation: https://github.com/apache/fineract/blob/develop/fineract-provider/src/main/java/org/apache/fineract/portfolio/self/account/data/SelfBeneficiariesTPTDataValidator.java#L90-L91
Update Beneficiary validation: https://github.com/apache/fineract/blob/develop/fineract-provider/src/main/java/org/apache/fineract/portfolio/self/account/data/SelfBeneficiariesTPTDataValidator.java#L126-L127

```
mysql> explain m_selfservice_beneficiaries_tpt;
+----------------+-------------+------+-----+---------+----------------+
| Field          | Type        | Null | Key | Default | Extra          |
+----------------+-------------+------+-----+---------+----------------+
   ... trimmed
| transfer_limit | bigint(20)  | YES  |     | 0       |                |
   ... trimmed
+----------------+-------------+------+-----+---------+----------------+

```

I tried tracking down the functional spec for this, but unfortunately it doesn't set out long/float - https://cwiki.apache.org/confluence/display/FINERACT/Customer+Self-Service+Phase+2

afk11:
For consideration - I have an alternative which uses long here, in case that's preferable https://github.com/openMF/mifos-mobile/pull/1815

PR #1925, opened 2023-02-24
raizlabs db error solved

### LINKED ISSUE #1924: Gradle build failed because of some dependency issue
**Summary:** 

no changes and mistakes during setup were from my side were done. The app did launch because of gradle build couldn't finish.

**Steps to reproduce:** 

Just setup the source code by usual steps and run the app in android studio

**Observed behavior:** 
image

adityarai004:
Fixes #1924

### DIFF
--- a/app/src/debug/java/org/mifos/mobile/api/BaseURL.kt
+++ b/app/src/debug/java/org/mifos/mobile/api/BaseURL.kt
@@ -7,7 +7,7 @@ package org.mifos.mobile.api
 class BaseURL {
     val url: String? = null
         get() = field
-                ?: PROTOCOL_HTTPS + API_ENDPOINT + API_PATH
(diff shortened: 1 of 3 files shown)


app/src/debug/java/org/mifos/mobile/api/BaseURL.kt
```diff
@@ -17,7 +17,7 @@ class BaseURL {
 
     companion object {
         const val API_ENDPOINT = "demo.mifos.community"
-        const val API_PATH = "/fineract-provider/api/v1/self/"
+        const val API_PATH = "/fineract-provider/api/v1x/"
```
jawidMuhammadi, line 20: Could you please clarify why you have changed the API_PATH?

adityarai004:
The end point of the API wasn't working while request was sending for example ID 
You can check it on postman, just try to send request with old and new API_PATH. Old one Throws error.

PR #1970, opened 2023-03-13
refactor: Removal of deprecated Kotlin-Android-Extensions Gradle Plugin

### LINKED ISSUE #1953: The 'kotlin-android-extensions' Gradle plugin is deprecated
While building the app it always shows a warning kotlin android extensions gradle plugin is deprecated.

Vaibhav2002:
Removed Kotlin-Android-Extensions and introduced ViewBinding.
ViewBinding had to be added in order to replace the long deprecated Kotlin-Synthetics included in Kotlin-Android-Extensions.
ViewBinding is the preferred way over Kotlin-Synthetics.
Kotlin Parcelize plugin had to be added in order to use Parcelization which was included in Kotlin-Android-Extensions.

Fixes #1953 

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,7 +9,7 @@ apply from: '../config/quality/quality.gradle'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
-    
+
     defaultConfig {
         applicationId "org.mifos.mobile"
         minSdkVersion rootProject.ext.minSdkVersion
@@ -44,7 +44,7 @@ android {
         }
     }
 
-    buildFeatures{
+    buildFeatures {
         viewBinding true
     }
 
@@ -117,7 +117,7 @@ dependencies {
     //Dagger dependencies
     kapt "com.google.dagger:dagger-compiler:$rootProject.daggerVersion"
     implementation "com.google.dagger:dagger:$rootProject.daggerVersion"
-    compileOnly 'javax.annotation:jsr250-api:1.0' 
+    compileOnly 'javax.annotation:jsr250-api:1.0'
     compileOnly 'com.github.pengrad:jdk9-deps:1.0'  //Required by Dagger2
 
     //Butter Knife
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,8 +2,8 @@ apply plugin: 'com.android.application'
 apply plugin: 'com.google.firebase.crashlytics'
 apply plugin: 'com.google.android.gms.oss-licenses-plugin'
 apply plugin: 'kotlin-android'
-apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
+apply plugin: 'kotlin-parcelize'
 
 apply from: '../config/quality/quality.gradle'
 
@@ -44,6 +44,10 @@ android {
         }
     }
 
+    buildFeatures{
+        viewBinding true
+    }
+
 
     sourceSets {
         def commonTestDir = 'src/commonTest/java'
@@ -71,10 +75,6 @@ android {
     lintOptions {
         abortOnError false
         disable 'InvalidPackage'
-    }
-
-    androidExtensions {
-        experimental = true
     }
 }
 
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
@@ -8,10 +8,9 @@ import androidx.lifecycle.ViewModelProviders
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.observers.DisposableObserver
 import io.reactivex.schedulers.Schedulers
-import kotlinx.android.synthetic.main.fragment_review_loan_application.*
-import kotlinx.android.synthetic.main.layout_error.*
 import okhttp3.ResponseBody
 import org.mifos.mobile.R
+import org.mifos.mobile.databinding.FragmentReviewLoanApplicationBinding
 import org.mifos.mobile.models.payload.LoansPayload
 import org.mifos.mobile.ui.activities.base.BaseActivity
 import org.mifos.mobile.ui.enums.LoanState
@@ -67,6 +66,9 @@ class ReviewLoanApplicationFragment : BaseFragment() {
 
     private lateinit var viewModel: ReviewLoanApplicationViewModel
 
+    private var _binding: FragmentReviewLoanApplicationBinding? = null
+    private val binding
+        get() = _binding!!
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
                               savedInstanceState: Bundle?): View? {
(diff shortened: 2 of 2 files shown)


Vaibhav2002:
@jawidMuhammadi can you please review this?

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
```diff
@@ -67,6 +66,9 @@ class ReviewLoanApplicationFragment : BaseFragment() {
 
     private lateinit var viewModel: ReviewLoanApplicationViewModel
 
+    private var _binding: FragmentReviewLoanApplicationBinding? = null
```
jawidMuhammadi, line 78: I would declare only one variable named `binding` using lateinit var. We don't require to have access to this view from outside of this fragment so it doesn't make any sense to me to have two private variables.

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
```diff
@@ -93,37 +95,43 @@ class ReviewLoanApplicationFragment : BaseFragment() {
...
-        tv_new_loan_application.text = viewModel.getLoanName()
-        tv_account_number.text = viewModel.getAccountNo()
-
-        btn_loan_submit.setOnClickListener {
+        _binding = FragmentReviewLoanApplicationBinding.bind(view)
+        setUpViews()
+
+    }
+
+    private fun setUpViews(): Unit = with(binding) {
```
jawidMuhammadi: I believe we don't need to make the return type of this function declared as `Unit`, it is already `Unit` by default so remove it please, and also use a `{` instead of `=`

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
```diff
@@ -132,22 +140,26 @@ class ReviewLoanApplicationFragment : BaseFragment() {
...
     }
 
-    fun showError(message: String?) = if (!Network.isConnected(activity)) {
-        iv_status.setImageResource(R.drawable.ic_error_black_24dp)
-        tv_status.text = getString(R.string.internet_not_connected)
-        ll_add_loan.visibility = View.GONE
-        ll_error.visibility = View.VISIBLE
-    } else {
-        Toaster.show(rootView, message)
+    fun showError(message: String?): Unit = with(binding) {
```
jawidMuhammadi: Same here

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
```diff
@@ -132,22 +140,26 @@ class ReviewLoanApplicationFragment : BaseFragment() {
...
+        val isConnected = Network.isConnected(activity)
+        if (!isConnected) {
+            Toaster.show(rootView, message)
+            return@with
+        }
+        llError.ivStatus.setImageResource(R.drawable.ic_error_black_24dp)
```
jawidMuhammadi: Would be better if you use the `apply` function  and set all the properties inside that function for `llErro`

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
```diff
@@ -132,22 +140,26 @@ class ReviewLoanApplicationFragment : BaseFragment() {
...
 
-    fun showError(message: String?) = if (!Network.isConnected(activity)) {
-        iv_status.setImageResource(R.drawable.ic_error_black_24dp)
-        tv_status.text = getString(R.string.internet_not_connected)
-        ll_add_loan.visibility = View.GONE
-        ll_error.visibility = View.VISIBLE
-    } else {
-        Toaster.show(rootView, message)
+    fun showError(message: String?): Unit = with(binding) {
+        val isConnected = Network.isConnected(activity)
```
jawidMuhammadi: I am not sure if you need the `isConnected` variable here

jawidMuhammadi (changes requested):

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
Vaibhav2002: Thats to avoid memory leaks, we should set binding to null in onDestroyView() which i completely forgot, will add this line.
The fragment lives even after its view is destroyed, so having a binding variable at that time will cause memory leak, so we usually clear it in onDestroyView and its automatically set in onViewCreated

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
Vaibhav2002: with() returns the the result from the lambda which is the last line, it may be Unit if the last line returns Unit or some other data type too. 
`inline fun <T, R> with(receiver: T, block: T.() -> R): R`
Also using = in place of { in this case makes the code less nested and look cleaner and more readable

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
Vaibhav2002: okay, will inline it in if condition

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
Vaibhav2002: Apply will work the same as With and also **as I said using `=` in place of `{` makes code much cleaner and readable due to less nesting**

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
jawidMuhammadi: That is a great point! Let's keep it then :)

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
jawidMuhammadi: It is a good practice that you are using `with()` but if the function doesn't suppose to return something then it is better to keep it as default. `=` looks cleaner and readable if a function returns something, not for Unit/Void :)

Vaibhav2002: If theres no issue with the extra nesting, then ill make it `{`

Vaibhav2002:
@jawidMuhammadi are the changes okay?

app/src/main/java/org/mifos/mobile/ui/fragments/ReviewLoanApplicationFragment.kt
```diff
@@ -132,27 +142,36 @@ class ReviewLoanApplicationFragment : BaseFragment() {
...
-        ll_add_loan.visibility = View.GONE
-        ll_error.visibility = View.VISIBLE
-    } else {
-        Toaster.show(rootView, message)
+    fun showError(message: String?) {
+        if (!Network.isConnected(activity)) {
+            Toaster.show(rootView, message)
+            return
+        }
+        binding.apply {
```
jawidMuhammadi: I was expecting something like:

```
binding.llError.apply {
      livStatus.setImageResource(R.drawable.ic_error_black_24dp)
      tvStatus.text = getString(R.string.internet_not_connected)
      root.visibility = View.VISIBLE
}
```

app/build.gradle
```diff
@@ -44,6 +44,10 @@ android {
         }
     }
 
+    buildFeatures{
```
jawidMuhammadi: There should be a space before the bracket, reformat your code, please

jawidMuhammadi (changes requested):

PratyushSingh07:
Already added viewbinding in the project @therajanmaurya , let's close it down

PR #1979, opened 2023-03-14
Home screen redesign

### LINKED ISSUE #1971: Unnecessary Deeply Nested Layout Hierarchy of fragment_home_ui.xml
**Summary:**

The layout of HomeFragment has very deep and nested Layout Hierarchy which is completely unnecessary and this makes the UI render very inefficient and adds load to the render thread. Making changes in views of this screen makes Android traverse through such a deeply nested ViewTree which makes it very time consuming

**Expected**

The Layout should be flattened using Constraint Layout to make the ViewTree less deep and improve Ui performance.

Vaibhav2002:
Deeply nested View Hierarchy is very bad for Ui Renderding
If any change happens in any view, RenderThread has to traverse such a deeply nested tree
This increases time take by a lot and causes frame drops.
Flattening view Heirarchy makes the height of this tree lesser and hence traversing time decreases, the flatter the tree, the faster the traversal.

Fixes #1971 

Both the layouts, the one after refactor and the one before looks the same, with slight styling improvements but with a huge difference in Layout Design and Performance

So the old Layout looked like:
image

And the new refactored Layout looks like:
image

You can clearly see how much the View Hierarchy has improved, this will get us huge performance gain while rendering this screen.

### DIFF
--- a/app/src/main/res/layout/fragment_home_ui.xml
+++ b/app/src/main/res/layout/fragment_home_ui.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.swiperefreshlayout.widget.SwipeRefreshLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.swiperefreshlayout.widget.SwipeRefreshLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/swipe_home_container"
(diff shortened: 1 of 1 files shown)


Vaibhav2002:
@jawidMuhammadi can you please review this PR

jawidMuhammadi:
It is good to use ConstraintLayout, especially for a complex screen, although I am not sure if this screen is counted as a complex one, and your design doesn't look similar to the current one, also, we are planning to use JetpacCompose so at this moment doing changes in UI might not be a good idea so for those reasons closing this PR and thanks for the effort.

Vaibhav2002:
Hi, I understand why this is being closed. Also the design doesn't look the same, coz i made improvements
But I would disagree that this Screen isn't complex, the design isn't complex i agree, but the way the XML was written, it was very very inefficient and made it complex. Writing such an XML for a not-so-complex screen and Constraint Layout should be preferred all the time, be it simple screen or complex, it has performance gains too

jawidMuhammadi:
@Vaibhav2002 Changing a design's look doesn't mean improvements, it just means redesigning some part of it. However, we could improve on how it has been implemented, if you strongly believe that the current XML file is insufficient, then please reopen the issue and create another PR but the design's look should not be changed because to change or alter the design we would need approval from the community members.

Vaibhav2002:
But 
> we are planning to use JetpackCompose so at this moment doing changes in UI might not be a good idea.

So should i send another PR or let it be and work on something else?
Also can you suggest me some issues to work on

PR #1984, opened 2023-03-14
fix:scrollable filters in landscape mode

### LINKED ISSUE #1980: Filters aren't scrollable in landscape mode
**Summary:**

- When the orientation of the app is changed to landscape mode the filters are no longer scrollable making it difficult for users to check the desired fields
- This issue is persistent even after changing the tabs to `Loan` and `Share`

**Steps to reproduce:**
- open the app
- click on accounts in the home section of the app
- in the top right corner , click on filters and then change the orientation

**Expected behaviour:**

The filters should have been scrollable even in the landscape mode

**Observed behaviour:**
- This is what you pretty much find if the orientation is changed to landscape
WhatsApp Image 2023-03-14 at 20 38 54

Device : Samsung M11
Android : 11

PratyushSingh07:
Fixes #1980

https://user-images.githubusercontent.com/90026952/225119065-deea8762-e821-4f03-8b64-d003d9aa023b.mp4

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
@@ -320,7 +320,7 @@ import javax.inject.Inject
         }
         wrapper.addView(checkBoxRecyclerView)
         if(resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE){
-            checkBoxRecyclerView!!.layoutParams.height = (getScreenHeight()/3.5).toInt()
+            checkBoxRecyclerView?.layoutParams?.height = (getScreenHeight()/3.5).toInt()
         }
         when (account) {
             AccountType.SAVINGS -> {
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
@@ -319,6 +319,9 @@ import javax.inject.Inject
             wrapper.setPadding(cardInnerPadding,cardInnerPadding,cardInnerPadding,cardInnerPadding)
         }
         wrapper.addView(checkBoxRecyclerView)
+        if(resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE){
+            checkBoxRecyclerView!!.layoutParams.height = (getScreenHeight()/3.5).toInt()
+        }
         when (account) {
             AccountType.SAVINGS -> {
                 if ((childFragmentManager.findFragmentByTag(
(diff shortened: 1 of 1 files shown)


PratyushSingh07:
@jawidMuhammadi could you please review this

app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
```diff
@@ -339,6 +328,9 @@ import javax.inject.Inject
                     checkBoxAdapter?.statusList = (childFragmentManager
                             .findFragmentByTag(getFragmentTag(0)) as AccountsFragment?)?.getCurrentFilterList()
                 }
+                if(resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE){
+                    checkBoxRecyclerView!!.layoutParams.height = (getScreenHeight()/3.5).toInt()
+                }
```
jawidMuhammadi: Is there any reason why you have copy pasted the same part of the code in 3 different places, what about if you remove this part of the code from the `when` block and maybe put it under line number 321? This way, you won't need to replicate it.

app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
```diff
@@ -359,6 +354,9 @@ import javax.inject.Inject
                     checkBoxAdapter?.statusList = (childFragmentManager
                             .findFragmentByTag(getFragmentTag(2)) as AccountsFragment?)?.getCurrentFilterList()
                 }
+                if(resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE){
+                    checkBoxRecyclerView!!.layoutParams.height = (getScreenHeight()/3.5).toInt()
```
jawidMuhammadi: Why divide by 3.5, could you please clarify?

jawidMuhammadi (changes requested):

app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
PratyushSingh07: 3.5 was need to adjust the size of the checkbox in landscape mode . This was tested on 3 devices 
- Samsung M11
- One plus nord 2
- Redmi note 9

app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
PratyushSingh07: You are right, we can write that at 321 and the if condition would be checked once but even if we write it 3 times even in this case the if condition would be executed once because at a time only one of the three account types would be clicked. 
Having said that , I can fix this and push it in the next commit if you would say so

jawidMuhammadi: Yeah, let's do that to avoid cod redundancy :)

PratyushSingh07: done !!

jawidMuhammadi: And the last change, can we use `?.` safe call operator instead of `!!`?

PratyushSingh07: done @jawidMuhammadi

jawidMuhammadi (approved):
Good Job! @PratyushSingh07

PR #1987, opened 2023-03-15
Moved presenters tests in presenters package! 

### LINKED ISSUE #1939: Test files not organized under package, making it  complicated  to add new tests
**Summary:**
 the test files for the "presenter" package are not organized under the "presenter" package name. 

**Steps to reproduce:**
- Navigate to   "app/src/test/java/org/mifos/mobile"   in the repository.
- You can see whole bunch of test files of  presenter package not organized under the package name presenter 

**Expected behavior:**

Test files should be organized under the same package as the source files, so that it's easy to add tests for other packages.

**Observed behavior:**

Test files  are not organized under the same package as the source files, and adding tests for other packages ( for example : api) would create a messy file structure.

Its actually related to the code structure and quality 

**Screenshots:**

image

deecodesss:
Fixes #1939

### DIFF
--- a/app/src/test/java/org/mifos/mobile/presenters/AccountOverviewPresenterTest.kt
+++ b/app/src/test/java/org/mifos/mobile/presenters/AccountOverviewPresenterTest.kt
@@ -1,4 +1,4 @@
-package org.mifos.mobile
+package org.mifos.mobile.presenters
 
 import android.content.Context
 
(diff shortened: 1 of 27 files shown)


PratyushSingh07:
We no longer need tests for presenters hence we will be pulling this down @therajanmaurya

PR #1988, opened 2023-03-15
Revamped the User Profile Fragment UI 

### LINKED ISSUE #1763: Enhancement in Existing UI- Profile screen
**Summary:**
The current UI needs to be revamped.
This image can be referred to.

deecodesss:
Fixes #1763 

image


PratyushSingh07:
This design isn't required @therajanmaurya as changing the UI of one single screen won't be ideal. Besides we are migrating to compose hence we can pull this down

PR #2001, opened 2023-03-16
fix : fixed the logic of comparator function that's used to sort notifications

### LINKED ISSUE #2000: In Utils the function that is used to compare notifications based on their timestamps is incorrect.
**Summary:**

In api -> local -> DatabaseHelper.kt, a comparator function utils -> NotificationComparator.kt -> compare() is used which is buggy and incorrect. 
In order to sort and store the notifications in descending order(based on their timestamp) a function compare is used which is incorrect and buggy(screenshot attached below)

**Steps to reproduce:**

How can we reproduce the issue?

check the function compare(in utils -> NotificationComparator.kt file)
attaching the screenshot below 

What did you expect the app to do?

comparator function should not have this bug, because due to this the working of notifications become prone to unintended errors.

What did you see instead?  Describe your issue in detail here.

Comparator function that's used to compare notifications has an error. 

What make and model device (e.g., Samsung Galaxy S3) did you encounter this on?  What Android
version (e.g., Android 4.0 Ice Cream Sandwich or Android 6.0 Marshmallow) are you running?  Is it
 the stock
version from the manufacturer or a custom ROM?

**Screenshots:**

Screenshot 2023-03-16 at 23 03 54

gururani-abhishek:
Fixes #2000 

Fixed the logic of comparator function that used to sort nofications(desc order) on the basis of their timestamps.

The comparator function that is called in DatabaseHelper.kt to sort the notifications in Descending order of their timestamp is incorrect and doesn't provide the right comparator function logic to sort objects in descending order. According to online java/kotlin documentation a comparator function should provide return values for all three cases.

1. updated the comparator function to provide return values for all three conditions(a > b ,a < b and a == b), where a and b are timestamps of any two notifications that are being compared.

### DIFF
--- a/app/src/main/java/org/mifos/mobile/utils/NotificationComparator.kt
+++ b/app/src/main/java/org/mifos/mobile/utils/NotificationComparator.kt
@@ -7,11 +7,11 @@ import java.util.*
  * Created by dilpreet on 14/9/17.
  */
 class NotificationComparator : Comparator<MifosNotification> {
-    override fun compare(mifosNotification1: MifosNotification, mifosNotification2: MifosNotification): Int {
+    override fun compare(firstNotification: MifosNotification, secondNotification: MifosNotification): Int {
         return when {
             // comparator function logic to sort notifications in the descending order of their timeStamp :
-            mifosNotification2.timeStamp < mifosNotification1.timeStamp -> -1
-            mifosNotification1.timeStamp < mifosNotification2.timeStamp -> 1
+            firstNotification.timeStamp > secondNotification.timeStamp -> -1
+            firstNotification.timeStamp < secondNotification.timeStamp -> 1
             else -> 0
         }
     }
--- a/app/src/main/java/org/mifos/mobile/utils/NotificationComparator.kt
+++ b/app/src/main/java/org/mifos/mobile/utils/NotificationComparator.kt
@@ -9,8 +9,9 @@ import java.util.*
 class NotificationComparator : Comparator<MifosNotification> {
     override fun compare(mifosNotification1: MifosNotification, mifosNotification2: MifosNotification): Int {
         return when {
+            // comparator function logic to sort notifications in the descending order of their timeStamp :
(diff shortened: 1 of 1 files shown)


gururani-abhishek:
Hey @jawidMuhammadi, please have a look :)
Further attaching an [online guide](https://clojure.org/guides/comparators) on Comparator Functions that more or less summaries my research on "Comparator Functions" that I did to solve this issue with our codebase.

app/src/main/java/org/mifos/mobile/utils/NotificationComparator.kt
```diff
@@ -9,9 +9,10 @@ import java.util.*
 class NotificationComparator : Comparator<MifosNotification> {
     override fun compare(mifosNotification1: MifosNotification, mifosNotification2: MifosNotification): Int {
         return when {
+            // comparator function logic to sort notifications in the descending order of their timeStamp :
             mifosNotification2.timeStamp < mifosNotification1.timeStamp -> -1
-            mifosNotification1.timeStamp > mifosNotification2.timeStamp -> 1
+            mifosNotification1.timeStamp < mifosNotification2.timeStamp -> 1
```
jawidMuhammadi: Great catch! However, both comparators are the same in line number 13 and 14, it is difficult to understand it for developers (although it's technically correct). Let's rename `mifosNotification1` variable to something like `firstNotification` and the other one to `secondNotificaiton`
then do the comparison like:
 `firstNotification.timpeStamp > secondNotificaiton.timeStamp` and 
`firstNotification.timpeStamp < secondNotificaiton.timeStamp `

jawidMuhammadi (changes requested):

app/src/main/java/org/mifos/mobile/utils/NotificationComparator.kt
gururani-abhishek: Thank you @jawidMuhammadi, I made the required changes : )

jawidMuhammadi (approved):
Looks Good To Me! :)

PR #2018, opened 2023-03-19
Nullable View receiver in HomeFragment & HomeOldFragment causing issue #1956

### LINKED ISSUE #1956: Nullable View receiver in HomeFragment & HomeOldFragment causing issue
**Summary:**

I'm experiencing an issue in the HomeFragment & HomeOldFragment of the Mifos Mobile app. Specifically, I'm seeing the error message "e: /home/runner/work/mifos-mobile/mifos-mobile/app/src/main/java/org/mifos/mobile/ui/fragments/HomeFragment.kt: (108, 14): Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type View?" when attempting to run the app.

From what I can gather, this seems to be related to a nullable receiver of type View that's being used in the HomeFragment. The error message suggests that only safe or non-null asserted calls are allowed on this type of receiver, but it doesn't seem like this is being followed in the code.

This is also the reason of the workflow failing on PR #1951 . Kindly refer [here](https://github.com/openMF/mifos-mobile/actions/runs/4393104320/jobs/7693291140)

oyehanif:
Fixes #1956 

Nullable View receiver in HomeFragment & HomeOldFragment causing issue.


app/src/main/res/values/strings.xml
```diff
@@ -588,6 +588,7 @@
     <string name="no_app_to_support_action">There is no application that support this action</string>
     <string name="theme">Theme</string>
     <string name="change_app_theme">Change the application theme</string>
+    <string name="TXT_PRODUCT_ERROR">Please Select Product Id</string>
```
jawidMuhammadi, line 591: Change this line of code to:
`  <string name="select_product_id">Please select product Id</string>`

jawidMuhammadi (changes requested):
Squash your commits and follow the commit and PR template styles, please

PratyushSingh07:
This issue no longer exists and hence can be pulled down

PratyushSingh07:
No such issue exists anymore, this can be pulled down @therajanmaurya

PR #2048, opened 2023-03-24
fix #1408 : fixing EditText error message on the login screen

### LINKED ISSUE #1408: Error in EditText View on Login Page doesn't get removed on entering data

gururani-abhishek:
Fixes #1408 

The error message is displayed in case of an error(eg : user left username/ or password blank and pressed login) But the error should go away when the user tries to input the username/ password again.

1. adding onTouch listener to both et_password and et_username, so when the user touches on these to input username/ or password clearUsernameError(), or clearPasswordError() is called and the error is dismissed.

2. tested on my pixel 2 and the fix runs smooth and fine.

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/activities/LoginActivity.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/activities/LoginActivity.kt
@@ -11,8 +11,10 @@ import androidx.appcompat.widget.AppCompatButton
 import butterknife.BindView
 import butterknife.ButterKnife
 import butterknife.OnClick
+import butterknife.OnTouch
 
 import com.google.android.material.textfield.TextInputLayout
+import kotlinx.android.synthetic.main.activity_login.*
 
 import org.mifos.mobile.R
 import org.mifos.mobile.models.payload.LoginPayload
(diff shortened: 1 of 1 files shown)


gururani-abhishek:
Hey @jawidMuhammadi, please have a look :)
This fix is also important as the concerning issue in a way also hampers with the UX, **because it hinders the users ability to hide or show password.** Not great for user experience, and something that should be solved on top priority. 

### Before the fix :

https://user-images.githubusercontent.com/59947871/227617807-4bf743a0-1faf-405b-9bc4-53df33dc61b5.mp4


### After the fix : 

https://user-images.githubusercontent.com/59947871/227617846-982e10bf-3228-4248-9b76-c08b9876798e.mp4

jawidMuhammadi (approved):
LGTM! 👍

PR #2092, opened 2023-04-06
fix: Make list view show correctly in dialog no matter the size of th…

### LINKED ISSUE #2091: long filter list not shown correctly in alert dialog, the buttons will be pushed down which will be invisible
**Summary:**

My Pixel XL API 30 (android 11) emulator will manifest this problem, long list in alert dialog (i.e. too much content in alert dialog) will cause the positive/neutral/negative buttons invisible (because there is no space left)

**Steps to reproduce:**

1) Launch the app and sign in (then finish any oobe steps)
2) When landed in Home-screen, click navigation-drawer and open left navigation view
3) Click "Accounts" navigation item, you will be directed to "Accounts" screen which has 3 tabs: Saving/Loan/Share
4) Click the top-right filter-icon to show the filter dialog (for each of the mentioned three tabs), the dialog will have problem in small size device (even in portrait mode)

**Expected behaviour:**

Dialog shows correctly

**Observed behaviour:**

Positive/Neutral/Negative buttons are invisible when filter-list is long

Pixel XL API 30 (android 11) emulator

**Screenshots:**
list_short_ok

list_long_not_ok

Mia0451:
…e device

Fixes #2091

After fix: 
long_list_after_fix
[after_fix.webm](https://user-images.githubusercontent.com/105191919/230300489-d9594658-4134-4db4-8e2d-f887f167117b.webm)

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/ClientAccountsFragment.kt
@@ -10,6 +10,7 @@ import android.view.*
 import android.widget.LinearLayout
 import android.widget.Toast
 import androidx.appcompat.widget.SearchView
+import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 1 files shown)


PratyushSingh07:
this issue no longer exists @therajanmaurya

PR #2095, opened 2023-04-06
fix : behavior of filters in loans tab

PratyushSingh07:
Fixes #2090 

https://user-images.githubusercontent.com/90026952/230488018-f09cf2fa-d547-4a1f-bb93-71f5079420c2.mp4

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2100, opened 2023-04-09
fix: logic to filter share accounts

### LINKED ISSUE #2099: Incorrect logic to filter share accounts
**Summary:**

image

In the above image the `else` block will get executed only when all of the below mentioned criterias are satisfied:
- accountsPresenter is not null 
- accountsPresenter?.getCheckedStatus(statusModelList) is null,
- filteredSavings.size is not equal to zero

However, this shouldn't be the case and the `else` block should be executed when filteredSavings.size is not equal to zero, irrespective of the outcome of the first 2 criteria

PratyushSingh07:
Fixes #2099 


PratyushSingh07:
hey @jawidMuhammadi can you review this as well as #2095 . Both of them had the same logical error which has been rectified

jawidMuhammadi (approved):
LGTM! :)

Issue #1397, opened 2020-03-09
AccountsFragment - applying no filter returns nothing


devansh-299:
**Summary:**

When the user presses Filter even without checking any checkbox the result returned is empty.
After referring to many other apps, I think in such situation the app should have either returned the same results as if the filter wasn't even applied.

**Steps to reproduce:**

* Go to Accounts
* Apply filter without selecting any checkbox for Savings/LoanAccounts

**Expected behaviour:**

The app should have returned the same results as if no filters were applied

**Observed behaviour:**

Empty list was displayed

**Device and Android version:**

XiaomiPocoF1
Android9
StockRom

**ScreenRecording:**

![GIF-200309_155913](https://user-images.githubusercontent.com/46667021/76205332-0294c080-6220-11ea-9443-f77645262a33.gif)

rutvik-panchal:
I would like to work on this issue

devansh-299:
> I would like to work on this issue

I am already working on it :sweat_smile:

rutvik-panchal:
I got the solution😅

devansh-299:
> I got the solutionsweat_smile

Actually I have already solved the issue just sending the PR
You can work on other open issues :)

ayush290104:
@PratyushSingh07 sir I opened a pr regarding this issue, can you review it and tell for any changes.

PR #2185, opened 2023-06-25
Filter button shows full list if no filter is selected in.

ayush290104:
#1397 
This pull request is in order to solve issue 1397 and from now selecting no filter gives the full list of accounts.

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
@@ -363,12 +363,19 @@ class AccountsFragment : BaseFragment(), OnRefreshListener, AccountsView {
      */
     fun filterSavingsAccount(statusModelList: List<CheckboxStatus?>?) {
         val filteredSavings: MutableList<SavingAccount?> = ArrayList()
-        if (accountsPresenter?.getCheckedStatus(statusModelList) != null && accountsPresenter != null) {
+        if (accountsPresenter?.getCheckedStatus(statusModelList) != null && accountsPresenter != null&&accountsPresenter?.getCheckedStatus(statusModelList)!!
+                .isNotEmpty()) {
             for (status in accountsPresenter?.getCheckedStatus(statusModelList)!!) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
```diff
@@ -363,14 +363,21 @@ class AccountsFragment : BaseFragment(), OnRefreshListener, AccountsView {
      */
     fun filterSavingsAccount(statusModelList: List<CheckboxStatus?>?) {
         val filteredSavings: MutableList<SavingAccount?> = ArrayList()
-        if (accountsPresenter?.getCheckedStatus(statusModelList) != null && accountsPresenter != null) {
+        if (accountsPresenter?.getCheckedStatus(statusModelList) != null && accountsPresenter != null&&accountsPresenter?.getCheckedStatus(statusModelList)!!
+                .isNotEmpty()) {
```
PratyushSingh07, line 367: use safe call operator instead of not-null assertion operator

PratyushSingh07 (changes requested):

PratyushSingh07:
you also need to create anothe branch , dont push the changes from the default branch . 
use `git checkout -b <branch-name>` to create a new branch 

It will be better if you could attach a video to demonstrate the changes

app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
ayush290104: Already we checked for null at the starting of the if statement so we can use a not null assertion.

ayush290104:
@PratyushSingh07 sir should i close this pr and and push from other branch in a different pr

app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
PratyushSingh07: you are right but using the safe call operator again provides an additional layer of null safety, ensuring that the subsequent method call is only executed if accountsPresenter is not null.

PR #2188, opened 2023-06-26
fix : no filter selected works properly in accounts fragment

ayush290104:
fix #1397 <on not selecting any filter account fragment now shows full list earlier it used to show empty list.>
before fix
https://github.com/openMF/mifos-mobile/assets/116818222/ec7f7395-9a28-462e-a9e8-e8e94063ef56
after fix

https://github.com/openMF/mifos-mobile/assets/116818222/92a09d8e-9c04-4d71-90b9-466e65c4801b

PR #2195, opened 2023-07-05
fix : no filter selected works properly in accounts fragment

### LINKED ISSUE #1397: AccountsFragment - applying no filter returns nothing

ayush290104:
fix #1397 
before fix
https://github.com/openMF/mifos-mobile/assets/116818222/ec7f7395-9a28-462e-a9e8-e8e94063ef56
after fix
https://github.com/openMF/mifos-mobile/assets/116818222/54c03f78-5404-48c1-abec-854f2feb3a01

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/AccountsFragment.kt
@@ -31,7 +31,7 @@ import org.mifos.mobile.utils.ComparatorBasedOnId
 import org.mifos.mobile.utils.Constants
 import org.mifos.mobile.utils.DividerItemDecoration
 import org.mifos.mobile.utils.Network
-import java.util.Collections
+import java.util.*
 import javax.inject.Inject
 
 /**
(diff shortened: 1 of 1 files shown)


ayush290104:
@jawidMuhammadi I did the required changes to pr #2188

PratyushSingh07:
hey @ayush290104 , it will be appreciated if you dont open a new PR everytime . Its recommened to have a separate branch for features that you are working on. In case you need to squash your commits then you can use the command `git rebase -i HEAD~N` where N is the number of commits that you need to squash

PR #2203, opened 2023-07-09
feat : updatepassword fragment mvvm migration

gururani-abhishek:
fixes : MOBILE-97

-> added viewModel and repository
-> refactored "RegistrationUiState" to "UiState" so that it could be used across registrationFragment and updatePasscode fragment 
-> refactored updatePassword fragment's exisiting code, to use more kotlin specific ways.

Fixes #Issue_Number

### DIFF
--- a/app/src/main/java/org/mifos/mobile/utils/UiState.kt
+++ b/app/src/main/java/org/mifos/mobile/utils/UiState.kt
@@ -1,0 +1,7 @@
+package org.mifos.mobile.utils
+
+sealed class UiState {
+    data class Error(val exception: Throwable) : UiState()
+    object Success : UiState()
+    object Loading : UiState()
+}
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/mifos/mobile/utils/UiState.kt
```diff
@@ -0,0 +1,7 @@
+package org.mifos.mobile.utils
+
+sealed class UiState {
```
jawidMuhammadi: It is better to give a specific name like `RegistrationUiState` then a general name

jawidMuhammadi (changes requested):
Other parts of the code looks good to me

app/src/main/java/org/mifos/mobile/utils/UiState.kt
gururani-abhishek: done plus resolved some conflicts.

jawidMuhammadi (approved):

PR #2257, opened 2023-08-05
refactor: AboutUs fragment migrated from xml to compose view

PratyushSingh07:
Jira ticket: https://mifosforge.jira.com/browse/MOBILE-136

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
<tr>
    <td></td>
    <td></td>
  </tr>
</table>


therajanmaurya (approved):

jawidMuhammadi:
@therajanmaurya Thank you for taking the time and reviewing this PR, this PR was a work in progress which is why I didn't have reviewed and I see recently it has been marked as Ready to Review. 

This PR is our initial integration of Jetpack compose and I believe we should start implementing it in a modular way and create a separate module of UI components so that we can use the same components through the App also @PratyushSingh07  works should act as a sample for other contributors. I have some change requests for this PR, therefore, I am reverting it back.

And also I am keen to know why you have implemented a different design than the old one @PratyushSingh07 .  I think the old one looks much more elegant.

app/build.gradle
```diff
@@ -82,6 +82,13 @@ android {
         viewBinding = true
     }
 
+    buildFeatures {
+        compose true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion "1.2.0-rc02"
```
jawidMuhammadi, line 89: I would prefer to use a stable version of this.

app/build.gradle
```diff
@@ -190,5 +197,12 @@ dependencies {
     implementation("com.google.dagger:hilt-android:2.44")
     kapt("com.google.dagger:hilt-android-compiler:2.47")
 
+    // Jetpack Compose
+    implementation "androidx.compose.material:material:1.3.0-rc01"
```
jawidMuhammadi, line 201: Add the versions in the `ext` block in the project level gradle file for consistency.

app/src/main/java/org/mifos/mobile/screens/AboutUsScreen.kt
```diff
@@ -0,0 +1,273 @@
...
+import org.mifos.mobile.ui.activities.PrivacyPolicyActivity
+import org.mifos.mobile.utils.Constants.licenseLink
+import org.mifos.mobile.utils.Constants.sourceCodeLink
+import org.mifos.mobile.utils.Constants.websiteLink
+import java.util.*
+
+
+@SuppressLint("UnusedMaterialScaffoldPaddingParameter")
+@Composable
+@Preview
```
jawidMuhammadi, line 36: Please remove the @Preview annotation from here and create another function to show a preview of this screen, that function should have a `Preview` suffix. For example, something like: `AboutViewPreview`

app/src/main/java/org/mifos/mobile/screens/AboutUsScreen.kt
```diff
@@ -0,0 +1,273 @@
...
+fun AboutView() {
```
jawidMuhammadi, line 37: Rename it to `AboutScreen`, because it is a screen-level composable.

PratyushSingh07:
@jawidMuhammadi theres no reason in particular but just that the new design looked better to me. I thought the card views would be better option as official website, licenses,source code and privacy policy are all clickable. Also , the color of the cards on the earlier designs looked worn out to me and hence I went with something a bit more bold

app/src/main/java/org/mifos/mobile/screens/AboutUsScreen.kt
```diff
@@ -0,0 +1,273 @@
...
+        context.getString(R.string.copy_right_mifos).replace("%1\$s", currentYear.toString())
+
+
+    AppTheme(isSystemInDarkTheme()) {
+        Scaffold {
+            Column(
+                modifier = Modifier
+                    .fillMaxSize()
+                    .background(MaterialTheme.colors.background)
+                    .padding(15.dp, 56.dp, 15.dp, 15.dp)
```
jawidMuhammadi, line 51: If all the values for padding are 15 then passing one value should be enough `padding(all=15.dp)`, I would use `16.dp` as it is mostly recommended material design value.

app/src/main/java/org/mifos/mobile/theme/Theme.kt
```diff
@@ -0,0 +1,46 @@
+package org.mifos.mobile.theme
+
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.darkColors
+import androidx.compose.material.lightColors
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Color
+
+private val LightThemeColors = lightColors(
```
jawidMuhammadi, line 9: I am not sure where you got all these color values, you might think of using https://material-components.github.io/material-components-android-compose-theme-adapter/ as we are in the initial phase of integration.

app/src/main/java/org/mifos/mobile/theme/Theme.kt
```diff
@@ -0,0 +1,46 @@
...
+    onSecondary = Color.White,
+    error = RedErrorLight,
+    background = BackgroundDark,
+    onBackground = Color.White,
+    surface = Black1,
+    onSurface = Color.White,
+)
+
+@Composable
+fun AppTheme(
```
jawidMuhammadi, line 38: Rename it to `MifosMobileTheme`

app/src/main/java/org/mifos/mobile/theme/Theme.kt
```diff
@@ -0,0 +1,46 @@
...
+    darkTheme: Boolean,
+    content: @Composable () -> Unit,
+) {
+    MaterialTheme(
+        colors = if (darkTheme) DarkThemeColors else LightThemeColors,
```
jawidMuhammadi, line 43: You might consider the dynamic color theming as well. This [tutorial](https://developer.android.com/codelabs/jetpack-compose-theming?continue=https%3A%2F%2Fdeveloper.android.com%2Fcourses%2Fpathways%2Fjetpack-compose-for-android-developers-2%23codelab-https%3A%2F%2Fdeveloper.android.com%2Fcodelabs%2Fjetpack-compose-theming#4) should help you to implement it.

app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
```diff
@@ -1,90 +1,41 @@
...
 */
 @AndroidEntryPoint
 class AboutUsFragment : BaseFragment() {
-    private var _binding: FragmentAboutUsBinding? = null
-    private val binding get() = _binding!!
 
-    private val licenseLink = "https://github.com/openMF/mifos-mobile/blob/development/LICENSE.md"
-    private val sourceCodeLink = "https://github.com/openMF/mifos-mobile"
-    private val websiteLink = "https://openmf.github.io/mobileapps.github.io/"
+    private lateinit var composeView: ComposeView
```
jawidMuhammadi, line 19: Do we really need this variable, is it possible to remove it?

app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
```diff
@@ -1,90 +1,41 @@
...
-            showWebsite()
-        }
-
-        binding.aboutLicensesContainer.setOnClickListener {
-            showOpenSourceLicenses()
-        }
-
-        binding.aboutPrivacyPolicyContainer.setOnClickListener {
-            showPrivacyPolicy()
+        composeView.setContent {
```
jawidMuhammadi, line 34: As we are using Compose with an existing view system, you should consider using `ViewCompositionStrategy`, for more details, take a look at [this document](https://developer.android.com/jetpack/compose/migrate/interoperability-apis/compose-in-views), please.

jawidMuhammadi (commented):
And please create a separate folder named `ui` in project level where we will put the common compose UI components/widgets

app/src/main/java/org/mifos/mobile/theme/Theme.kt
PratyushSingh07: we are using a mifos ui library , I picked up some of the values from there and the remaing from a github repository

jawidMuhammadi:
- Create a data class named `AboutUsItem(name, iconUrl?)`, You can get the list of items from a ViewModel -if we don't have one, create it, please.
- Create a composable component named as `AboutUsItemCard`, where you should pass the `name`, `iconId`, and a `callback method` (Use lambda) for handling user clicks. If the iconId is null then don't show an icon.
- Column is fine to use, however, you could use LazyColumn as well and use the `AboutUsItemCard` for each item

jawidMuhammadi:
@PratyushSingh07, I am okay with your new design. Please open a new PR and consider the above change request. Let me know if you have further questions. Take your time researching and don't rush to meet the GSoC deadline. You can continue working on this after GSoC as well if you would like to.

app/src/main/java/org/mifos/mobile/screens/AboutUsScreen.kt
PratyushSingh07: all values aren't 15 here, but I will make sure to use 16 instead

PratyushSingh07:
> * Create a data class named `AboutUsItem(name, iconUrl?)`, You can get the list of items from a ViewModel -if we don't have one, create it, please.
> * Create a composable component named as `AboutUsItemCard`, where you should pass the `name`, `iconId`, and a `callback method` (Use lambda) for handling user clicks. If the iconId is null then don't show an icon.
> * Column is fine to use, however, you could use LazyColumn as well and use the `AboutUsItemCard` for each item

I am sorry @jawidMuhammadi but I dont understand the use of viewmodel in this particular case.

app/build.gradle
PratyushSingh07: we are using kotlin version `1.6.21` in the project and hence we need this version of `kotlinCompilerExtensionVersion `
image

We can either continue with this or we would have to upgrade the kotlin version . Although upgrading the kotlin version may come with problem of its own

app/src/main/java/org/mifos/mobile/ui/fragments/AboutUsFragment.kt
PratyushSingh07: yes, i will use `ComposeView` directly

jawidMuhammadi:
Well, you could create the list of `AboutUsItem` in the fragment (UI) as well, but creating a list counted as business logic, so it would be better if you separate the UI logic from business/data logic. Maybe in the future, requirements changes and we want the list to be dynamic and managed by the backend. So it is always a good practice to take scalability and consistency in mind. And also helps to follow the [Unidirectional Data Flow](https://developer.android.com/jetpack/compose/architecture#udf) architecture in your compose implementation.  I hope it makes sense to you.

app/build.gradle
jawidMuhammadi: Ah, okaay, let's use this version temporarily then.

PratyushSingh07:
I have done all the changes that you had asked for except the viewmodel one, I will try to implement that as well @jawidMuhammadi

PR #2262, opened 2023-08-06
feat: implemented coroutine in recent transaction viewmodel

PratyushSingh07:
Jira ticket: https://mifosforge.jira.com/browse/MOBILE-118


app/src/test/java/org/mifos/mobile/viewModels/RecentTransactionViewModelTest.kt
```diff
@@ -108,10 +116,12 @@ class RecentTransactionViewModelTest {
             RecentTransactionUiState.EmptyTransaction,
             viewModel.recentTransactionUiState.value
         )
+        Dispatchers.resetMain()
```
jawidMuhammadi: Instead of resetting the dispatcher in every function, you can create a coroutine rule and use it in each test class. You can [follow this tutoria](https://medium.com/swlh/unit-testing-with-kotlin-coroutines-the-android-way-19289838d257)l to achieve it.

jawidMuhammadi (changes requested):

app/src/test/java/org/mifos/mobile/viewModels/RecentTransactionViewModelTest.kt
PratyushSingh07: thank you jawid, I have done the changes, do review them and let me know if they are correct. If yes , then I would replicate the same in the existing PRs

jawidMuhammadi (approved):
LGTM!

PR #2267, opened 2023-08-08
feat: implemented coroutines in notification viewmodel

PratyushSingh07:


PratyushSingh07:
used coroutine tests rule here

Issue #2361, opened 2023-10-10
Text in snackbar isnt visible when i try to change the password


PratyushSingh07:
**Summary:**

When I try to change the password , appropriate text should have been visible in the snackbar

**Steps to reproduce:**

https://github.com/openMF/mifos-mobile/assets/90026952/679f352b-5403-4821-ae61-68e102f282b5

**Expected behaviour:**

Text inside the snackbar should have been visible. 

**Device and Android version:**

Device: Pixel 7
Android: 13

PratyushSingh07:
Please check if the text is visible in light theme as well before submitting a PR

SekhGulamMainuddin:
Hi @PratyushSingh07 I would like to work on this isuue

oyehanif:
@PratyushSingh07 i want to work on this issue.

PratyushSingh07:
This issue has already been solved in PR #2364

PR #2364, opened 2023-10-12
fixes: snackbar error messages on registration and password change

### LINKED ISSUE #2363: Bug: Json displayed in snackbars instead of a correct error message
**Summary:**
When creating an account or changing password from Settings, the snackbar message displayed is not proper and does not effectively communicate the error/cause of the error. I would like to amend this.

Summarize your issue in one sentence (what goes wrong, what did you expect to happen).

**Steps to reproduce:**

How can we reproduce the issue?

**Expected behaviour:**

What did you expect the app to do?

**Observed behaviour:**

What did you see instead?  Describe your issue in detail here.

Samsung galaxy M31 API 30.

What make and model device (e.g., Samsung Galaxy S3) did you encounter this on?  What Android
version (e.g., Android 4.0 Ice Cream Sandwich or Android 6.0 Marshmallow) are you running?  Is it
 the stock
version from the manufacturer or a custom ROM?

**Screenshots:**
Screenshot_20231011-132044_Mifos Mobile_previous
Screenshot_20231011-122041_Mifos Mobile

Can be created by pressing the Volume Down and Power Button at the same time on Android 4.0 and higher.

srishti-R:
Fixes #2363 #2361 

Previously the entire JSON of the error was shown in snackbar conveying nothing meaningful. I have extracted the actual error message that comes within the error body and that is displayed as an error message.
Screenshot_20231011-133540_Mifos Mobile
Screenshot_20231011-134520_Mifos Mobile
Screenshot_20231011-134604_Mifos Mobile
Screenshot_20231012-182914_Mifos Mobile

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/RegistrationFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/RegistrationFragment.kt
@@ -100,7 +100,7 @@ class RegistrationFragment : BaseFragment() {
 
                 is RegistrationUiState.Error -> {
                     hideProgress()
-                    showError(state.exception)
+                    showError(getString(state.exception))
                 }
             }
         }
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/UpdatePasswordFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/UpdatePasswordFragment.kt
@@ -63,7 +63,7 @@ class UpdatePasswordFragment : BaseFragment(), TextWatcher, OnFocusChangeListene
 
                 is RegistrationUiState.Error -> {
                     hideProgress()
-                    showError(state.exception)
+                    showError(getString(state.exception))
                 }
             }
         }
--- a/app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
+++ b/app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
@@ -7,6 +7,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.launch
+import org.mifos.mobile.R
 import org.mifos.mobile.repositories.UserAuthRepository
 import org.mifos.mobile.utils.RegistrationUiState
 import javax.inject.Inject
@@ -67,8 +68,7 @@ class RegistrationViewModel @Inject constructor(private val userAuthRepositoryIm
             if (response?.isSuccessful == true) {
                 _registrationUiState.value = RegistrationUiState.Success
             } else {
-                _registrationUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                _registrationUiState.value = RegistrationUiState.Error(R.string.could_not_register_user_error)
             }
         }
     }
@@ -81,8 +81,7 @@ class RegistrationViewModel @Inject constructor(private val userAuthRepositoryIm
                 _registrationVerificationUiState.value =
                     RegistrationUiState.Success
             } else {
-                _registrationVerificationUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                _registrationVerificationUiState.value = RegistrationUiState.Error(R.string.could_not_register_user_error)
             }
         }
     }
--- a/app/src/main/java/org/mifos/mobile/viewModels/UpdatePasswordViewModel.kt
+++ b/app/src/main/java/org/mifos/mobile/viewModels/UpdatePasswordViewModel.kt
@@ -6,6 +6,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.launch
+import org.mifos.mobile.R
 import org.mifos.mobile.repositories.ClientRepository
 import org.mifos.mobile.repositories.UserAuthRepository
 import org.mifos.mobile.utils.RegistrationUiState
@@ -35,20 +36,16 @@ class UpdatePasswordViewModel @Inject constructor(
         viewModelScope.launch {
             _updatePasswordUiState.value = RegistrationUiState.Loading
             val response = userAuthRepositoryImp.updateAccountPassword(newPassword, confirmPassword)
-
             try {
                 if (response?.isSuccessful == true) {
                     _updatePasswordUiState.value = RegistrationUiState.Success
                     clientRepositoryImp.updateAuthenticationToken(newPassword)
                 } else {
-                    _updatePasswordUiState.value =
-                        response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                    _updatePasswordUiState.value = RegistrationUiState.Error(R.string.could_not_update_password_error)
                 }
             } catch (e: Throwable) {
-                _updatePasswordUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                _updatePasswordUiState.value = RegistrationUiState.Error(R.string.could_not_update_password_error)
             }
         }
     }
-
 }
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -646,5 +646,6 @@
     </string-array>
     <string name="app_info">App Info</string>
     <string name="login_failed">Login Failed, Please Try Again Later.</string>
-
+    <string name="could_not_update_password_error">We were unable to update password.</string>
+    <string name="could_not_register_user_error">We were unable to register the user.</string>
 </resources>
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/RegistrationVerificationFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/RegistrationVerificationFragment.kt
@@ -51,7 +51,7 @@ class RegistrationVerificationFragment : BaseFragment() {
 
                 is RegistrationUiState.Error -> {
                     hideProgress()
-                    showError(state.exception)
+                    showError(getString(state.exception))
                 }
             }
         }
--- a/app/src/main/java/org/mifos/mobile/utils/RegistrationUiState.kt
+++ b/app/src/main/java/org/mifos/mobile/utils/RegistrationUiState.kt
@@ -1,7 +1,7 @@
 package org.mifos.mobile.utils
 
 sealed class RegistrationUiState {
-    data class Error(val exception: String) : RegistrationUiState()
+    data class Error(val exception: Int) : RegistrationUiState()
     object Success : RegistrationUiState()
     object Loading : RegistrationUiState()
 }


app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
```diff
@@ -82,7 +87,11 @@ class RegistrationViewModel @Inject constructor(private val userAuthRepositoryIm
                     RegistrationUiState.Success
             } else {
                 _registrationVerificationUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                    response?.errorBody()?.string()?.let {
+                        val errorResponseObject = JSONObject(it)
+                        val error = errorResponseObject.getString("error")
+                        RegistrationUiState.Error(error)
+                    }
```
PratyushSingh07: why not create a function instead of writing the same code twice ? Also what is the use of the hardcoded string "error" on line 92?

app/src/main/java/org/mifos/mobile/viewModels/UpdatePasswordViewModel.kt
```diff
@@ -35,18 +36,30 @@ class UpdatePasswordViewModel @Inject constructor(
...
+                        }
                 }
             } catch (e: Throwable) {
                 _updatePasswordUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                    response?.errorBody()?.string()?.let {
+                        val errorResponseObject = JSONObject(it)
+                        val error = errorResponseObject.getString("error")
+                        RegistrationUiState.Error(error)
+                    }
```
PratyushSingh07: create a function here as well

app/src/main/java/org/mifos/mobile/ui/fragments/UpdatePasswordFragment.kt
```diff
@@ -63,6 +65,10 @@ class UpdatePasswordFragment : BaseFragment(), TextWatcher, OnFocusChangeListene
 
                 is RegistrationUiState.Error -> {
                     hideProgress()
+                    if (TextUtils.isEmpty(state.exception)) {
+                        showError(getString(R.string.not_found_error))
+                        return@observe
+                    }
```
PratyushSingh07: if I update the value of live data to `R.string.not_found_error` in case of an error then will I not be able to observe them later? why do we need to handle this explicitly in our ui ?

PratyushSingh07 (changes requested):

PratyushSingh07:
Please change the commit message from `fixes: snackbar error messages on registration and password change` to `fixes #2363 : snackbar error messages on registration and password change`

PratyushSingh07:
> Fixes #2363

Also I see that this PR also solves issue #2361 . Add this issue as well in the PR description

app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
srishti-R: error is the field that comes in json, I wanted to extract that value to be shown in snackbar. There's no point putting it in strings.xml because this will never be shown to any user.

app/src/main/java/org/mifos/mobile/ui/fragments/UpdatePasswordFragment.kt
srishti-R: I tried doing it that way but there is no context object available inside the ViewModel class. From the StackOverflow answer [here]( https://stackoverflow.com/questions/51451819/how-to-get-context-in-android-mvvm-viewmodel#:~:text=you%20can%20access%20the%20application,resources%2C%20preferences%2C%20etc..&text=Finally%20i%20got%20the%20easiest,context%20in%20viewModel%20using%20MVVM.) it is not a good practice to pass activity context in view model class. And since I do not have a context object, I can't add `getString(R.string.not_found_error)` in `RegistrationUiState.Error(exception)` inside ViewModel class, hence handling this here.

PratyushSingh07: I don't think you need to pass any context in the first place. Updating the value incase of error should be enough

app/src/main/java/org/mifos/mobile/viewModels/UpdatePasswordViewModel.kt
srishti-R: made the changes :)

app/src/main/java/org/mifos/mobile/ui/fragments/UpdatePasswordFragment.kt
srishti-R: um, I think there is some misunderstanding here. This block exclusively handles the case where we are checking 404 error and no error from API is coming. So need to show this string in UI. Otherwise I would have taken the error from json.

PratyushSingh07: I understand you are trying to handle 404 code, but what I was thinking was to change the type from String to Int in the RegistrationUiState sealed class. If we do say we can just update the value of live data to R.string.not_found_error and it would look Something like this : 
```
  _updatePasswordUiState.value = RegistrationUiState.Error(R.string.not_found_error)
```

In this way in our ui we will just need to show the error as  `showError(getString(state.exception))`. Thus we wont have to use TextUtils and JsonObject classes

What do you think ?

srishti-R: For that I will have to change the signature of RegistrationUiState.Error from string to int, then in my `extractErrorFromErrorBodyResponse` method how will I put the error which is coming in json as string? Or did you mean I put this generic error for any kind of error that's coming?

PratyushSingh07: Yes, put relevant errors in else and catch block. They should be enough for our use case. (We can get the code from response to display more relevant messages 🤔)

You won't be needing your function anymore if you do this

srishti-R: made the changes.

app/src/main/java/org/mifos/mobile/ui/fragments/RegistrationFragment.kt
```diff
@@ -265,7 +265,7 @@ class RegistrationFragment : BaseFragment() {
         )
     }
 
-    fun showError(msg: String?) {
+    fun showError(msg: Int) {
```
PratyushSingh07: you werent supposed to change the signature of these functions. You were supposed to handle them in the  ui in a way i described.

app/src/main/java/org/mifos/mobile/ui/fragments/UpdatePasswordFragment.kt
```diff
@@ -152,10 +152,10 @@ class UpdatePasswordFragment : BaseFragment(), TextWatcher, OnFocusChangeListene
 
     }
 
-    fun showError(message: String?) {
+    fun showError(message: Int) {
         var errorMessage = message
         if (!Network.isConnected(activity)) {
-            errorMessage = getString(R.string.no_internet_connection)
+            errorMessage = R.string.no_internet_connection
```
PratyushSingh07: revert the type to String for these functions and just handle them where they are being observed

app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
```diff
@@ -7,6 +7,8 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.launch
+import org.json.JSONObject
```
PratyushSingh07: we no longer need this import

app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
```diff
@@ -67,8 +69,13 @@ class RegistrationViewModel @Inject constructor(private val userAuthRepositoryIm
             if (response?.isSuccessful == true) {
                 _registrationUiState.value = RegistrationUiState.Success
             } else {
-                _registrationUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                response?.code()?.compareTo(400)?.let {
+                    if (it >= 0) {
+                        _registrationUiState.value = RegistrationUiState.Error(R.string.could_not_register_user_error)
+                    }
+                    return@launch
```
PratyushSingh07: why are we using launch here ?

app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
```diff
@@ -81,8 +88,13 @@ class RegistrationViewModel @Inject constructor(private val userAuthRepositoryIm
                 _registrationVerificationUiState.value =
                     RegistrationUiState.Success
             } else {
-                _registrationVerificationUiState.value =
-                    response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                response?.code()?.compareTo(400)?.let {
```
PratyushSingh07: do we actually need this comparison ? we are anyway showing a generic error .

app/src/main/res/values/strings.xml
```diff
@@ -646,5 +646,6 @@
     </string-array>
     <string name="app_info">App Info</string>
     <string name="login_failed">Login Failed, Please Try Again Later.</string>
-
+    <string name="could_not_update_password_error">Could not update password. Please try again later.</string>
```
PratyushSingh07: Its a long string, keep it concise. How about `We were unable to update password` ?

app/src/main/res/values/strings.xml
```diff
@@ -646,5 +646,6 @@
...
+    <string name="could_not_register_user_error">Could not register user. Please try again later.</string>
```
PratyushSingh07: Same here , keep it concise .

app/src/main/java/org/mifos/mobile/viewModels/UpdatePasswordViewModel.kt
```diff
@@ -6,6 +6,8 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.launch
+import org.json.JSONObject
```
PratyushSingh07: remove unused imports

app/src/main/java/org/mifos/mobile/viewModels/UpdatePasswordViewModel.kt
```diff
@@ -35,20 +37,22 @@ class UpdatePasswordViewModel @Inject constructor(
...
             val response = userAuthRepositoryImp.updateAccountPassword(newPassword, confirmPassword)
-
             try {
                 if (response?.isSuccessful == true) {
                     _updatePasswordUiState.value = RegistrationUiState.Success
                     clientRepositoryImp.updateAuthenticationToken(newPassword)
                 } else {
-                    _updatePasswordUiState.value =
-                        response?.errorBody()?.string()?.let { RegistrationUiState.Error(it) }
+                    response?.code()?.compareTo(400)?.let {
```
PratyushSingh07: same here, we can remove the comparison

PratyushSingh07 (changes requested):

app/src/main/java/org/mifos/mobile/viewModels/RegistrationViewModel.kt
srishti-R: To handle the edge case when no response is coming but code is 404, so that 2 popups are not shown

PratyushSingh07: If we remove the comparisons with 400 then we can skip it entirely

srishti-R: changes done

PratyushSingh07:
LGTM @srishti-R , great job

PratyushSingh07:
@therajanmaurya review and merge this as well

PR #2399, opened 2023-10-15
Fix: #2397 Added System Language Option

### LINKED ISSUE #2068: App theme doesn't change when we change the main mobile theme
**Summary:**

App theme remains set to the one defined in the settings section and doesn't change with the main mobile theme

**Steps to reproduce:**

Try to change the main mobile theme and then see the behaviour

**Expected behaviour:**

App theme should have changed wrt the main mobile theme

**Observed behaviour:**

https://user-images.githubusercontent.com/90026952/227758208-8f6e153a-88db-48c0-a54a-115151ca3088.mp4

Device : Samsung M11
Version : 11

### LINKED ISSUE #2397: System Default Language option not present in Settings language dialog
**Summary:**

There is no option for System Default Language in the Settings screen of the app,

**Steps to reproduce:**

Home Screen->Open The Navigation Menu from left side->Open Settings->Click on Language

**Expected behaviour:**

Needs to have an option for System Default Language in Choose Language Dialog on the Settings Screen.

**Observed behaviour:**

No option for System Default Language in Choose Language Dialog on the Settings Screen.

Android Emulator: Pixel 7 Pro (Android 14 - UpsideDownCake)

**Screenshots:**

<p>

</p>

SekhGulamMainuddin:
Fixes #2397

Video Recordings:
**System Language Option Added**

https://github.com/openMF/mifos-mobile/assets/73953395/af150639-0add-4211-840e-3f327b7f5d61

**If System Language is Not Present then English is Default**
Here I switched to Punjabi which is not supported by the App so it is showing English by Default.

https://github.com/openMF/mifos-mobile/assets/73953395/e75cfc44-ebb4-4c19-9de2-6358d35ec63f

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/fragments/SettingsFragment.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/fragments/SettingsFragment.kt
@@ -21,6 +21,7 @@ import org.mifos.mobile.utils.ConfigurationDialogFragmentCompat
 import org.mifos.mobile.utils.ConfigurationPreference
 import org.mifos.mobile.utils.Constants
 import org.mifos.mobile.utils.LanguageHelper
+import java.util.Locale
 
 /**
  * Created by dilpreet on 02/10/17.
@@ -132,7 +133,23 @@ class SettingsFragment : PreferenceFragmentCompat(), OnSharedPreferenceChangeLis
     override fun onSharedPreferenceChanged(p0: SharedPreferences?, p1: String?) {
         val preference = findPreference(p1)
         if (preference is ListPreference) {
-            LanguageHelper.setLocale(context, preference.value)
+            val isSystemLanguage =
+                (preference.value == resources.getStringArray(R.array.languages_value)[0])
+            prefsHelper.putBoolean(
+                context?.getString(R.string.default_system_language),
+                isSystemLanguage
+            )
+            if (!isSystemLanguage) {
+                LanguageHelper.setLocale(context, preference.value)
+            } else {
+                if (!resources.getStringArray(R.array.languages_value)
+                        .contains(Locale.getDefault().language)
+                ) {
+                    LanguageHelper.setLocale(context, "en")
+                } else {
+                    LanguageHelper.setLocale(context, Locale.getDefault().language)
+                }
+            }
             val intent = Intent(activity, activity?.javaClass)
             intent.putExtra(Constants.HAS_SETTINGS_CHANGED, true)
             startActivity(intent)
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -535,6 +535,7 @@
     </string-array>
 
     <string name="language_type" translatable="false">language_type</string>
+    <string name="default_system_language" translatable="false">default_system_language</string>
     <string name="theme_type" translatable="false">theme_type</string>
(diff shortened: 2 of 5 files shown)


SekhGulamMainuddin:
@PratyushSingh07 can you please review my PR

PratyushSingh07:
Any PR should solve a single issue. Are these two issues dependent on each other? If not then please segregate them

SekhGulamMainuddin:
> Any PR should solve a single issue. Are these two issues dependent on each other? If not then please segregate them

Actually removing this single function solved the issue for theme but it was unnecessary for System Language also that is the reason it solved both the issues.

```
    override fun attachBaseContext(base: Context) { 
         super.attachBaseContext(onAttach(base, Locale.getDefault().language)) 
     }
```

app/src/main/java/org/mifos/mobile/ui/fragments/SettingsFragment.kt
```diff
@@ -132,7 +134,23 @@ class SettingsFragment : PreferenceFragmentCompat(), OnSharedPreferenceChangeLis
     override fun onSharedPreferenceChanged(p0: SharedPreferences?, p1: String?) {
         val preference = findPreference(p1)
         if (preference is ListPreference) {
-            LanguageHelper.setLocale(context, preference.value)
+            val isSystemLanguage =
+                (preference.value == resources.getStringArray(R.array.languages_value)[0])
+            val preferences = PreferenceManager.getDefaultSharedPreferences(context)
+            preferences.edit()
+                .putBoolean(context?.getString(R.string.default_system_language), isSystemLanguage)
+                .apply()
```
PratyushSingh07: we already have a PreferenceHelper class in our project, i dont think we will need this

app/src/main/res/values/strings.xml
```diff
@@ -620,6 +622,7 @@
     </string-array>
 
     <string-array name="languages_value" translatable="false">
+        <item>System_Language</item>
```
PratyushSingh07, line 625: do we really need to add this option if we are changing the language dynamically? I mean the moment we set our mobile to follow a particular language it should be reflected in our app at that very instance.

PratyushSingh07 (changes requested):
Please format your code before you push it here. You are inconsistent with the formatting

app/src/main/res/values/strings.xml
SekhGulamMainuddin: Yes, we need it because if we are dependent on the System's Language then having a System Language value will be helpful. And sometimes if the System's Language is not supported in our App then English is made as Default. I have attached the videos for it also on the top.

This is also similar to the app theme preference on System Theme, Dark Theme, and Light Theme.

app/src/main/java/org/mifos/mobile/ui/fragments/SettingsFragment.kt
SekhGulamMainuddin: Changed to PreferenceHelper.

SekhGulamMainuddin:
> Please format your code before you push it here. Your are inconsistent with the formatting

@PratyushSingh07 Sorry for it. I formatted the code now and added the changes as you said. Please review it.

SekhGulamMainuddin:
@PratyushSingh07 can you please re-review this?

PR #2550, opened 2024-03-26
safe way to collect flows from android UI

MuindiStephen:
The safest way to collecting flows from UI eg. HomeActivity

repeatOnLifecycle is a suspend function that takes a [Lifecycle.State](https://developer.android.com/reference/android/arch/lifecycle/Lifecycle.State) as a parameter which is used to automatically create and launch a new coroutine with the block passed to it when the lifecycle reaches that state, and cancel the ongoing coroutine that’s executing the block when the lifecycle falls below the state

### DIFF
--- a/app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
+++ b/app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
@@ -2,6 +2,7 @@ package org.mifos.mobile.utils
 
 import android.graphics.Bitmap
 import org.mifos.mobile.models.client.Client
+import org.mifos.mobile.ui.user_profile.UserDetails
 
 
 sealed class UserDetailUiState {
@@ -10,3 +11,11 @@ sealed class UserDetailUiState {
     data class ShowUserDetails(val client: Client) : UserDetailUiState()
     data class ShowUserImage(val image: Bitmap?) : UserDetailUiState()
 }
+
+/**
+data class UserDetailUiStates (
+    var isLoading:Boolean? = false,
+    var error: String? = null,
+    var allUserDetails: List<Client> = listOf()
+)
+ */
--- a/app/src/main/java/org/mifos/mobile/models/User.kt
+++ b/app/src/main/java/org/mifos/mobile/models/User.kt
@@ -11,5 +11,5 @@ data class User(
     var isAuthenticated: Boolean = false,
     var username: String? = null,
     var base64EncodedAuthenticationKey: String? = null,
-    var permissions: List<String> = ArrayList(),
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
```diff
@@ -2,6 +2,7 @@ package org.mifos.mobile.utils
 
 import android.graphics.Bitmap
 import org.mifos.mobile.models.client.Client
+import org.mifos.mobile.ui.user_profile.UserDetails
```
MuindiStephen, line 5: remove unused imports

app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
```diff
@@ -10,3 +11,11 @@ sealed class UserDetailUiState {
     data class ShowUserDetails(val client: Client) : UserDetailUiState()
     data class ShowUserImage(val image: Bitmap?) : UserDetailUiState()
 }
+
```
MuindiStephen, line 14: Ignore this class, just that its another way to declare ui state

app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
```diff
@@ -10,3 +11,11 @@ sealed class UserDetailUiState {
...
+/**
+data class UserDetailUiStates (
+    var isLoading:Boolean? = false,
+    var error: String? = null,
+    var allUserDetails: List<Client> = listOf()
+)
+ */
```
therajanmaurya, line 21: remove this unused code block.

therajanmaurya (changes requested):

app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
MuindiStephen: > remove this unused code block.

Yes. Looking into that.

app/src/main/java/org/mifos/mobile/utils/UserDetailUiState.kt
```diff
@@ -2,6 +2,7 @@ package org.mifos.mobile.utils
 
 import android.graphics.Bitmap
 import org.mifos.mobile.models.client.Client
+import org.mifos.mobile.ui.user_profile.UserDetails
 
 
 sealed class UserDetailUiState {
```
MuindiStephen, line 8: To be removed, since unused block of code stmt.

therajanmaurya:
We are not using fragments anymore.

PR #2559, opened 2024-03-29
fix: currency display symbol null

### LINKED ISSUE #2433: some currencies are not visible in the accounts section
**Summary:**

- some currencies are not visible in the accounts section namely in the saving and loan section

**Steps to reproduce:**

- open the accounts section in the home screen

**Expected behaviour:**

- currency should have been visible for all of the transacitons

**Observed behaviour:**

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

Device: Pixel 7
Android : 14

AbrahamOsmondE:
Fixes #2433

Fix for Albanian LEK, previously showing null:

**Before:** 
Screenshot 2024-03-30 at 1 17 40 AM

**After:** 
Screenshot 2024-03-30 at 1 12 30 AM

### DIFF
--- a/app/src/main/java/org/mifos/mobile/ui/adapters/ClientChargeAdapter.kt
+++ b/app/src/main/java/org/mifos/mobile/ui/adapters/ClientChargeAdapter.kt
@@ -40,7 +40,7 @@ class ClientChargeAdapter(
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
         val charge = getItem(position)
-        var currencyRepresentation = charge?.currency?.displaySymbol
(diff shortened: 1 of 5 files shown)


AbrahamOsmondE:
The list of currency are taken from a DB Sql, whose implementation defers from tenant to tenant. An the default currency list and properties can be found [here](https://github.com/apache/fineract/blob/5fa2295bd0923916d9553717f2936d422fd04fef/fineract-db/multi-tenant-demo-backups/bk_mifostenant_default.sql#L734)

It seems that MySQL does not support some of these currency symbols ([source](https://gist.github.com/voskobovich/537b2000108e4781f70b)), causing `displaySymbol` to be `null` for some currencies.

We should instead implement a fallback for these cases

PR #2590, opened 2024-05-17
fix#2589: Migrate QrCodeReaderFragment from XML to compose

### LINKED ISSUE #2589: refactor: Migrate QrCodeReaderFragment from XML to compose

itsPronay:
fixes #2589 

[Screencast_05_17_2024_03:42:28_PM.webm](https://github.com/openMF/mifos-mobile/assets/78101731/4f77e27f-fce4-4d16-9062-05c253a40a33)


app/src/main/java/org/mifos/mobile/ui/qrCodeReader/QrCodeScreen.kt
```diff
@@ -0,0 +1,90 @@
...
+                Text(text = stringResource(id = R.string.add_beneficiary))
+            }
+        }
+    ) {
+        Box(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(it),
+        ) {
+            AndroidView({ scannerView })
```
AvneetSingh2001, line 58: dont we have any other options than AndroidView?

AvneetSingh2001 (changes requested):

app/src/main/java/org/mifos/mobile/ui/qrCodeReader/QrCodeScreen.kt
itsPronay: "Unfortunately, as of now, there's no direct replacement in Jetpack Compose for ZXingScannerView that doesn't use AndroidView." 

I looked up online and it was the only way I found. Probably we don't have another way of doing it.

PR #2604, opened 2024-06-11
refactor: notification screen migration to compose

AvneetSingh2001:
Jira Task: [MM-7](https://mifosforge.jira.com/browse/MM-7)

[MM-7]: https://mifosforge.jira.com/browse/MM-7?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/org/mifos/mobile/models/notification/MifosNotification.kt
```diff
@@ -31,8 +31,8 @@ class MifosNotification : BaseModel() {
         this.timeStamp = timeStamp
     }
 
-    fun isRead(): Boolean? {
-        return read
+    fun isRead(): Boolean {
+        return read ?: false
```
therajanmaurya, line 35: Good job, Kotlin is built for handling nullable

app/src/main/java/org/mifos/mobile/ui/notification/NotificationFragment.kt
```diff
@@ -0,0 +1,54 @@
...
+        (activity as? BaseActivity)?.hideToolbar()
+        return mifosComposeView(requireContext()) {
+            NotificationScreen(
+                navigateBack = { activity?.onBackPressed() }
+            )
+        }
+    }
+
+    companion object {
+        @JvmStatic
```
therajanmaurya: We use `@JvmStatic` if we call this function from java class but I don't think that is the case here so we can remove it.

therajanmaurya (changes requested):

app/src/main/java/org/mifos/mobile/ui/notification/NotificationFragment.kt
AvneetSingh2001: updated.

PR #2724, opened 2024-12-24
fix: Fix visibility issues in Saving Account Withdraw screen

HekmatullahAmin:
Changed background color to make fields visible in dark mode.

Screenshot 2024-12-24 at 10 56 43
Screenshot 2024-12-24 at 15 41 13

### DIFF
--- a/feature/savings/src/main/java/org/mifos/mobile/feature/savings/savingsAccountWithdraw/SavingsAccountWithdrawScreen.kt
+++ b/feature/savings/src/main/java/org/mifos/mobile/feature/savings/savingsAccountWithdraw/SavingsAccountWithdrawScreen.kt
@@ -76,7 +76,11 @@ private fun SavingsAccountWithdrawScreen(
 ) {
     val context = LocalContext.current
 
-    Column(modifier = modifier.fillMaxSize()) {
+    Column(
+        modifier = modifier
+            .fillMaxSize()
+            .background(MaterialTheme.colorScheme.background),
+    ) {
         MifosTopBar(
             navigateBack = { navigateBack(false) },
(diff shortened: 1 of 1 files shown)


feature/savings/src/main/java/org/mifos/mobile/feature/savings/savingsAccountWithdraw/SavingsAccountWithdrawScreen.kt
```diff
@@ -76,7 +76,11 @@ private fun SavingsAccountWithdrawScreen(
 ) {
     val context = LocalContext.current
 
-    Column(modifier = modifier.fillMaxSize()) {
+    Column(
+        modifier = modifier
+            .fillMaxSize()
+            .background(MaterialTheme.colorScheme.background),
```
niyajali, line 82: Don't apply color directly on Column instead change it on `MifosTopBar` or configure theme and it will be reflected in all screens

niyajali (changes requested):

feature/savings/src/main/java/org/mifos/mobile/feature/savings/savingsAccountWithdraw/SavingsAccountWithdrawScreen.kt
HekmatullahAmin: Thank you for the feedback! I understand your concern about theming, and it’s definitely something that could be improved across the app. However, the team's current focus seems to be less on how the app looks and more on other priorities, as there are broader UI issues to address. My main goal here was to quickly fix the visibility issue while keeping consistency with how other screens are implemented. I appreciate your input and will keep theming in mind for future updates!

Issue #185, opened 2017-03-15
Improve UX implementation of Loan Account Details


karsai1993:
Summary:
The UI implementation of LoanAccountsDetailFragment should be improved. 

Steps to reproduce:
1, After login, open LOAN tab
2, Click one of them e.g.: 000000500

Expected behavior:
The textViews should be placed in the center of their layout to have a better look on them. Additionally, in case of the value of Next Installment, there should be the currency shown as it is in case of Outstanding Balance.

Observed behavior:
TextView values are not in the center of their layout. E.g. in case of Next Installment and its value: they were too close to each other (see attached image). Value of Next Installment does not shown the currency.

Device and Android version:
Samsung Ace 3
API 17 - 4.2.2

Screenshots:
![load_account_details_wrong_ui](https://cloud.githubusercontent.com/assets/15123875/23967313/b48a7d38-09be-11e7-9087-851882167f76.png)

karsai1993:
I have already implemented the solution. Please take a look at the attach image showing the result of my implementation. I am ready to commit and send a pull request.
![improved_load_account_details_wrong_ui](https://cloud.githubusercontent.com/assets/15123875/23967371/f231e450-09be-11e7-8f44-7ff5ea28b45e.png)

karsai1993:
The currency symbol should be after the value, not before it. See attached image:
![newest_improved_load_account_details_wrong_ui](https://cloud.githubusercontent.com/assets/15123875/23975603/0c3899a2-09e2-11e7-9d9d-35e00ba25007.png)

dilpreet2028:
Fixed in #332

PR #2779, opened 2025-02-19
feat(feature:accounts): Migrate accounts module to CMP

HekmatullahAmin:
Fixes - [Jira-#185](https://mifosforge.jira.com/browse/MM-185)

https://github.com/user-attachments/assets/40db28da-0e91-452b-8fdb-c6c9dc41b684

https://github.com/user-attachments/assets/b994d95c-fb2d-43a8-90a2-b5f4329f19de

https://github.com/user-attachments/assets/00b1c471-bf84-4ab4-8ff3-afb4182cc0e1

### DIFF
--- a/feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/component/AccountFilterCheckbox.kt
+++ b/feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/component/AccountFilterCheckbox.kt
@@ -68,9 +68,10 @@ internal fun AccountFilterCheckBox(
 private fun AccountCheckBoxItem(
     checkboxStatus: CheckboxStatus,
     onCheckedChange: (Boolean) -> Unit,
+    modifier: Modifier = Modifier,
 ) {
     Row(
-        modifier = Modifier
+        modifier = modifier
             .fillMaxWidth()
             .padding(horizontal = 12.dp, vertical = 4.dp),
         verticalAlignment = Alignment.CenterVertically,
--- a/feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/component/AccountFilterCheckbox.kt
+++ b/feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/component/AccountFilterCheckbox.kt
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * See https://github.com/openMF/mobile-mobile/blob/master/LICENSE.md
+ */
+package org.mifos.mobile.feature.accounts.component
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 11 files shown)


HekmatullahAmin:
@therajanmaurya , @niyajali  

I’ve completed the multi-module structure for the account feature, separating it into accounts (base), savings-account, loan-account, and share-account modules. Each module is now independent and managed via the accounts module for navigation and display.

✅ What I’ve Done:

Modularized the account feature into separate modules.

KDoc documentation added to ViewModels, StatusUtils, and FilterUtil for better clarity.

Refactored functions: Moved necessary functions to their respective modules and made them more concise.

Renamed properties and parameters for better readability.

Improved the filter dialog: It now dynamically displays checkboxes based on the Status property of each account, making it more structured.

Added cards for each account entry to enhance UI presentation.

Moved module-specific strings to their respective modules for better organization.

🛠️ To Be Worked on Later:

AccountTypeItemIndicator is used across all three modules inside the card. It would be better to move this to the design system module so all three account modules can use it from there.

UseCase Implementation:

I haven’t added UseCases yet because my focus was first on making the ViewModels more readable.

Adding UseCases right now would mostly introduce extra layers without much added benefit, but I’m open to discussing this further.

📌 Next Steps:

Please review my PRs in the following order:

MM-188

MM-182

MM-183

MM-185

If everything looks good, kindly merge them in this order.

After merging, please delete the old account module, as it's no longer needed after modularization.

feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/component/AccountFilterCheckbox.kt
```diff
@@ -0,0 +1,94 @@
...
+                },
+            )
+        }
+    }
+}
+
+@Composable
+private fun AccountCheckBoxItem(
+    checkboxStatus: CheckboxStatus,
+    onCheckedChange: (Boolean) -> Unit,
```
niyajali, line 70: This composable doesn't contain default modifier,I wonder how does detekt checks passed..

feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/utils/StatusUtils.kt
```diff
@@ -0,0 +1,112 @@
...
+        )
+    }
+
+    /**
+     * Retrieves the list of available checkbox statuses for share accounts.
+     *
+     * @return A list of [CheckboxStatus] representing different share account statuses.
+     */
+    internal fun getShareAccountCheckboxes(): List<CheckboxStatus> {
+        return arrayListOf(
```
niyajali: Use `listOf()` builder function instead of arrayList here and above

niyajali (changes requested):

feature/accounts/src/commonMain/kotlin/org/mifos/mobile/feature/accounts/component/AccountFilterCheckbox.kt
HekmatullahAmin: @niyajali I’ve made the suggested changes by replacing arrayListOf() with listOf() and adding the missing default modifier.

Regarding why Detekt checks passed, I believe it is because when we use private and internal access modifiers, Detekt does not flag it as an issue. However, I think this should be improved so that Detekt also enforces the same rule for private and internal access modifiers to ensure consistency.

niyajali (approved):

PR #2783, opened 2025-02-25
feat(feature:recent-transaction): migrate to CMP

biplab1:
Fixes - [Jira-#192](https://mifosforge.jira.com/browse/MM-192)

* Web app keeps showing the loading animation after login

Didn't create a Jira ticket, click [here](https://mifosforge.jira.com/jira/software/c/projects/MM/issues/) to create new.

Please Add Screenshots If there are any UI changes.

| Before                                     | After                                  |
|--------------------------------------------|----------------------------------------|
|                                            |                                        |

https://github.com/user-attachments/assets/f823f2c6-0ae9-4be2-bf85-a87d2edcf02e

https://github.com/user-attachments/assets/dc9d8c7c-69a5-4dcf-97d4-6371fc9193c0

https://github.com/user-attachments/assets/7d229cc0-5d8c-41e4-8713-5b1c8f234e22

https://github.com/user-attachments/assets/34526674-38f9-4d5c-a2bf-e89f646f4e6d

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

- [ ] If you have multiple commits please combine them into one commit by squashing them.

PR #2810, opened 2025-03-12
feat: [feature/recent _transactions] to CMP

itsPronay:
Fixes - https://mifosforge.jira.com/browse/MM-192

### This PR resolves the inconsistent loading wheel on PR - #2783 by @biplab1

https://github.com/user-attachments/assets/9345924e-6d68-469a-8edb-ab65b3d766a3

- [x] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything


feature/recent-transaction/src/commonMain/kotlin/org/mifos/mobile/feature/recent/transaction/screen/RecentTransactionScreen.kt
```diff
@@ -208,60 +209,29 @@ private fun RecentTransactionListItem(
...
                     modifier = Modifier
                         .weight(1f)
                         .alpha(0.7f),
-                    color = MaterialTheme.colorScheme.onSurface,
                 )
                 Text(
-                    text = DateHelper.getDateAsString(transaction?.submittedOnDate),
+                    text = DateHelper.getDateAsString(transaction!!.submittedOnDate),
                     style = MaterialTheme.typography.labelMedium,
                     modifier = Modifier.alpha(0.7f),
```
niyajali: remove this alpha values here and above

niyajali (dismissed):
@itsPronay @biplab1 why icon color doesn't reflect based on theme mode, and we're not showing **account no** and **Icon and amount** it's not properly aligned, show account no with masked text (show only the last few digits) and enhance the UI

itsPronay:
@niyajali we were using icon from `painterresource `[ which was originally black] and that's why it was just showing black. I have added tint, so its visible properly now.
image

About the alignment issue,  it is perfectly alligned, I added demo data to test before, where i didn't add` type`, that's why it used to look like that. 

and we can't show `accountNo` either since it is not a field in `transaction`
image

feature/recent-transaction/src/commonMain/kotlin/org/mifos/mobile/feature/recent/transaction/screen/RecentTransactionScreen.kt
```diff
@@ -90,13 +91,15 @@ private fun RecentTransactionScreen(
     onRefresh: () -> Unit,
     loadMore: (offset: Int) -> Unit,
     modifier: Modifier = Modifier,
+    viewModel: RecentTransactionViewModel = koinViewModel(),
 ) {
-    val context = LocalContext.current
+    val isNetworkAvailable by viewModel.isNetworkAvailable.collectAsStateWithLifecycle()
+
```
HekmatullahAmin: you are initializing two times viewmodel just do it once in the top hierarchy and pass the isNetworkAvailable to below composable

feature/recent-transaction/src/commonMain/kotlin/org/mifos/mobile/feature/recent/transaction/viewmodel/RecentTransactionViewModel.kt
```diff
@@ -0,0 +1,99 @@
...
+        _isRefreshing.value = true
+        loadInitialTransactions()
+    }
+
+    fun loadPaginatedTransactions(offset: Int) {
+        _isPaginating.value = true
+        loadRecentTransactions(clientId, offset, limit)
+    }
+
+    fun loadInitialTransactions() {
```
HekmatullahAmin, line 66: just for the reason to assigning the state to Loading you created another function. you can set the state inside the load recentTransactions and get rid of this.

feature/recent-transaction/src/commonMain/kotlin/org/mifos/mobile/feature/recent/transaction/screen/RecentTransactionScreen.kt
itsPronay: @HekmatullahAmin Thanks for pointing it out. I didn't check biplab's code. gonna fix it

feature/recent-transaction/src/commonMain/kotlin/org/mifos/mobile/feature/recent/transaction/viewmodel/RecentTransactionViewModel.kt
itsPronay: This function[which is also getting called from the screen] was created so that we can call `loadInitialTransactions()`  from viewmodel and pass `clientId` and `limit` as well. [we can't access clientId and limit from the screen since its private and this function passes those values into `loadInitialTransactions`]

therajanmaurya (approved):

PR #2860, opened 2025-07-16
feat: home ui & viewModel

Nagarjuna0033:
Fixes - [Jira-#268](https://mifosforge.jira.com/browse/MM-268)

|Current Design                    | Figma Design                         |
|  |


Nagarjuna0033:
server response is getting unauthorized Will update image

Nagarjuna0033:
@niyajali have a doubt if the client don't have any loan account then what we can show in place of the card ( that blue color card ), asked this with Saksham maybe he is busy. Currently I am getting server response as unauthorized in this case I am not getting anu client accounts. so what to show instead of that card

niyajali:
Show the loan apply card for example "apply for home loan" with some details and a button to apply

Nagarjuna0033:
> Show the loan apply card for example "apply for home loan" with some details and a button to apply

Ok @niyajali but what to show if we get unauthorized error from server side in place of card

niyajali:
@Nagarjuna0033 I'm not talking about as per the current scenario but it can be done that way when user doesn't have any loan and as for current scenario make all the necessary changes and when we're able to login test everything

revanthkumarJ (approved):

Nagarjuna0033:
> @Nagarjuna0033 I'm not talking about as per the current scenario but it can be done that way when user does have any loan and as for current scenario make all the necessary changes and when we're able to login test everything

@niyajali made necessary changes and after we able to login will test everything

Nagarjuna0033:
@niyajali could you please merge this pr

niyajali (approved):

PR #2861, opened 2025-07-19
feat: passcode ui & viewModel

Nagarjuna0033:
Fixes - [Jira-#269](https://mifosforge.jira.com/browse/MM-269)

| Figma | Design |
|--------|---------|
| image | image  |

https://github.com/user-attachments/assets/ffcbbd76-88d9-417d-8aa2-d6cfa1770987


revanthkumarJ (approved):

niyajali:
@Nagarjuna0033 Lock icon look bigger than figma UI/UX, and fix the static analysis check

Nagarjuna0033:
> @Nagarjuna0033 Lock icon look bigger than figma UI/UX, and fix the static analysis check

@niyajali updated

Nagarjuna0033:
> @Nagarjuna0033 Lock icon look bigger than figma UI/UX, and fix the static analysis check

@niyajali updated

cmp-navigation/src/commonMain/kotlin/cmp/navigation/authenticated/AuthenticatedNavigation.kt
```diff
@@ -42,6 +47,16 @@ internal fun NavGraphBuilder.authenticatedGraph(
         notificationDestination(
             navigateBack = navController::popBackStack,
         )
+
+        statusDestination(
+            navigateToDestination = {
+                if (it == "login") {
```
niyajali: Aren't we following type-safe navigation @Nagarjuna0033

niyajali (changes requested):
Could you clarify why we're doing like this?

cmp-navigation/src/commonMain/kotlin/cmp/navigation/authenticated/AuthenticatedNavigation.kt
Nagarjuna0033: > Aren't we following type-safe navigation @Nagarjuna0033

@niyajali We are following type safe the reason why I am doing like this is from the status screen we are navigating to different screens like after successful login we are navigating status screen to login and when it is passcode we are navigating to home so to detect the next route I am passing a string so that we can navigate to respective screen.

If it is login we need to popup the entire back stack, so I am checking the condition and navigating to LoginRoute

feature/passcode/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+            implementation(compose.components.resources)
+            implementation(compose.components.uiToolingPreview)
+            implementation(libs.kotlinx.serialization.json)
+        }
+    }
+}
+
+dependencies {
+    debugImplementation(compose.uiTooling)
+}
```
HekmatullahAmin: either use this dependencies and inside it `debugImplementation(compose.uiTooling)` or `implementation(compose.components.uiToolingPreview)` inside the commonMain. it is extra. so u can keep this `dependencies{}` block

feature/passcode/src/commonMain/kotlin/org/mifos/mobile/feature/passcode/PasscodeViewModel.kt
```diff
@@ -0,0 +1,146 @@
...
+                PasscodeMode.Confirm -> {
+                    val confirm = passcodeBuilder.toString()
+
+                    if (confirm == state.firstPasscode) {
+                        viewModelScope.launch {
+                            userPreferencesRepository.setPasscode(confirm)
+                            sendEvent(
+                                PasscodeEvent.OnPasscodeConfirm(
+                                    eventType = EventType.SUCCESS.name,
+                                    eventDestination = "unlocked",
```
HekmatullahAmin: it would be good to have constants for this type of values so if we change it in one place other place also have the change.

feature/passcode/src/commonMain/kotlin/org/mifos/mobile/feature/passcode/PasscodeScreen.kt
```diff
@@ -0,0 +1,339 @@
...
+    onBackspaceClick: () -> Unit,
+    onSendClick: () -> Unit,
+    isSendEnabled: Boolean,
+    modifier: Modifier = Modifier,
+) {
+    val keys = listOf(
+        listOf("1", "2", "3"),
+        listOf("4", "5", "6"),
+        listOf("7", "8", "9"),
+        listOf("Backspace", "0", "Send"),
```
HekmatullahAmin: Replace `Backspace`, `Send` with Sealed Class. in that case your will `when{}` block will also look tidy and definite.

```
sealed class Key {
    data class Digit(val number: String) : Key()
    object Backspace : Key()
    object Send : Key()
}
```

```
val layout = listOf(
        listOf(PasscodeKey.Digit("1"), PasscodeKey.Digit("2"), PasscodeKey.Digit("3")),
        listOf(PasscodeKey.Digit("4"), PasscodeKey.Digit("5"), PasscodeKey.Digit("6")),
        listOf(PasscodeKey.Digit("7"), PasscodeKey.Digit("8"), PasscodeKey.Digit("9")),
        listOf(PasscodeKey.Backspace, PasscodeKey.Digit("0"), PasscodeKey.Send)
    )
```

```
row.forEach { key ->
                    when (key) {
                        is PasscodeKey.Digit -> DigitKeyButton()
                        PasscodeKey.Backspace -> KeyButton()
                        PasscodeKey.Send -> KeyButton()
                    }
                }
```

Nagarjuna0033: Thanks @HekmatullahAmin

cmp-navigation/src/commonMain/kotlin/cmp/navigation/authenticated/AuthenticatedNavigation.kt
Nagarjuna0033: @niyajali if this is good could you merge this pr

Nagarjuna0033: @niyajali any change needed

niyajali: I didn't understand what you're trying to say but I'm merging this for now and will revisit later. And create a ticket for this. @Nagarjuna0033

niyajali (approved):

PR #291, opened 2017-04-07
feat:adding double tap on back button to exit feature on home screen …

ghost:
…#177

Fixes #{Issue Number}

Please Add Screenshots If any UI changes.

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `MifosStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2873, opened 2025-07-28
feat: Payfrom DropDown Card UI

revanthkumarJ:
Fixes - [Jira-#291](https://mifosforge.jira.com/browse/MM-291)

Figma
image
image

Design

https://github.com/user-attachments/assets/84d4fab7-df4b-4fa3-9524-5a285b785d25


Nagarjuna0033:
@revanthkumarJ can we select the account after dropdown if yes it should reflect in the card? record and upload video

Nagarjuna0033:
@revanthkumarJ this looks good and add some animation while dropdown like fade in or fadeout some other

revanthkumarJ:
@Nagarjuna0033 i have updated video check once.

On selecting from options it should reflect in card and also calls a unit function updating selected account in parent component.
Regarding animation i tried but for this design animations not looking good, the current version where options coming down as drop down and going up looks good to me.

Issue #297, opened 2017-04-21
Status color of closed savings account should be black in stead of blue


santoshconflux:
In web demo server:
![demo](https://cloud.githubusercontent.com/assets/23022667/25274117/47bd34ce-26ac-11e7-8269-448c578e74f4.png)

In mobile app:

![ss4](https://cloud.githubusercontent.com/assets/23022667/25274148/63b0d21c-26ac-11e7-8733-173f1bb4268a.jpg)

PR #2882, opened 2025-07-31
fix : Make Filter Checkbox Title Clickable Horizontally (Not Just Checkbox)

revanthkumarJ:
Fixes - [Jira-#297](https://mifosforge.jira.com/browse/MM-297)

https://github.com/user-attachments/assets/714a6bca-59e8-45d6-b15b-b0f41a18bd85


Nagarjuna0033:
@revanthkumarJ can you confirm you are using the same checkboxes that we are used in savings account filters seems different

Issue #299, opened 2017-05-19
Show message and retry button if app failed to fetch LoanTemplate while applying for loan


therajanmaurya:
**Summary:**

In Present we are fetching the loan template and inflating into the UI content but if app failed to fetch the loan template. It's can make app crash or HTTP request failure reason. It's better to show a **message and retry** button if app failed to fetch loan template.

**Steps to reproduce:**

Try to apply loan application without internet connection. App is crashinng.

**Expected behavior:**

It should show the message and retry button at the centre of screen

**Observed behavior:**

It showing empty UI and on click apply loan. Application is crashing.

**Device and Android version:**

MOTO G Turbo 3

PR #2883, opened 2025-07-31
refactor: integrate savings transfer & deposit flow

Nagarjuna0033:
Fixes - [Jira-#299](https://mifosforge.jira.com/browse/MM-299)


Nagarjuna0033:
@revanthkumarJ I have integrated savings transfer and deposit flow but getting some issue with Api, you have done this before so you will have better idea can you pull this and fix that one thing

PR #2925, opened 2025-08-11
feat: Comment settings unused options

revanthkumarJ:
Fixes - [Jira-#338](https://mifosforge.jira.com/browse/MM-338)

Profile commented - Because no apis available for self user to edit details
Theme commented - Because current design system colours won't support dark theme
Endpoint commented - Because no ui/ux screens present for it


feature/settings/src/commonMain/kotlin/org/mifos/mobile/feature/settings/settings/SettingsScreen.kt
```diff
@@ -66,18 +67,7 @@ internal fun SettingsScreen(
...
-                    SettingsItems.AppInfo,
-                    SettingsItems.AuthPasscode,
-                    SettingsItems.Language,
-                    SettingsItems.FAQ,
-                    SettingsItems.Password,
-                    -> navigateToScreen.invoke(events.item)
-
-                    else -> {}
-                }
+                navigateToScreen.invoke(events.item)
```
Nagarjuna0033, line 70: @revanthkumarJ may be it's better if we use when for safety

Nagarjuna0033 (changes requested):

PR #2937, opened 2025-08-15
chore: remove unused modules

Nagarjuna0033:
Fixes - [Jira-#380](https://mifosforge.jira.com/browse/MM-380)


revanthkumarJ (changes requested):
remove about,help,user-profile and other too which are not used

Nagarjuna0033:
> remove about,help,user-profile and other too which are not used

Yes Removing

PR #421, opened 2017-09-23
fix: Title bar now shows the correct title.

mayank8318:
Fixes #412 

Please Add Screenshots If any UI changes.
21730725_1813167445378728_7820823189008678912_n

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [x] Run the unit tests with `./gradlew check` to make sure you didn't break anything

- [x] If you have multiple commits please combine them into one commit by squashing them.

PR #2970, opened 2025-09-11
feat(home): add empty state and account creation flow

Nagarjuna0033:
Fixes - [Jira-#421](https://mifosforge.jira.com/browse/MM-421)

This commit introduces an empty state for the home screen when the user has no accounts.
It also adds a bottom sheet for account creation and refactors the `MifosDashboardCard` to support this new flow.
The `MifosProgressIndicator` now loops indefinitely.
The registration flow now includes the middle name.

image
image


therajanmaurya:
@Nagarjuna0033 Make changes 

image

1. This top header should not change and keep it in older way.



image

1.  Remove white background card
2. Make boarder dotted line
3. Use small button instead normal button 



image

1. Reorder the account list and make savings account first and loan second 

**Fix checks**

PR #3030, opened 2025-12-25
Adding centralized unauthorized handling on 401 responses

Kartikey15dem:
### Fixes
- Jira: [MM-466](https://mifosforge.jira.com/jira/software/c/projects/MM/issues/MM-466)

### Overview
This PR ensures the app performs a clean logout and returns the user to the Auth/Login flow whenever the backend responds with `401 Unauthorized`.

<p align="center">
  401 Unauthorized logout flow
</p>

Unauthorized handling is centralized in the network layer via a Ktor interceptor callback (`onUnauthorized`). When a 401 is detected, the interceptor triggers `preferencesRepository.logOut()`. Since `authState` is derived from stored preferences, it automatically updates to `Unauthenticated`, and the root navigation reacts by switching to the Auth route.

### Key Changes
- Added/updated Ktor interceptor support for unauthorized handling via a suspend callback:
  - `onUnauthorized = suspend { preferencesRepository.logOut() }`
- Used the existing preference-driven `authState` to propagate logout state across the app (no extra event mechanism required).
- Updated `RootNavViewModel` to react to `AuthState.Unauthenticated` and navigate to the Auth/Login route.
- Home screen edge case: added a lightweight session validation API call when the Home screen enters composition/resumes so a stale/expired session can still trigger a 401 and force logout even if no other API calls are made.

### Technical Flow
401 Response -> Ktor interceptor -> `onUnauthorized` callback -> `preferencesRepository.logOut()` -> `authState` emits `Unauthenticated` -> Root navigation switches to `Auth`

[MM-466]: https://mifosforge.jira.com/browse/MM-466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ.

## Summary by CodeRabbit

* **New Features**
  * App now detects invalid authentication and routes users to the auth flow automatically.
  * On-resume and on-composition authentication checks run to verify session validity.
  * Network client detects 401 responses and triggers a graceful logout.

* **Bug Fixes**
  * Avoids redundant authorization checks during ongoing network-change processing.

### DIFF
--- a/cmp-navigation/src/commonMain/kotlin/cmp/navigation/rootnav/RootNavViewModel.kt
+++ b/cmp-navigation/src/commonMain/kotlin/cmp/navigation/rootnav/RootNavViewModel.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
 import org.mifos.mobile.core.data.repository.UserDataRepository
 import org.mifos.mobile.core.datastore.model.AppSettings
 import org.mifos.mobile.core.model.AuthState
@@ -27,6 +28,16 @@ class RootNavViewModel(
 ) {
 
     init {
+        viewModelScope.launch {
+            userDataRepository.authState
+                .collect { authState ->
+                    if (authState is AuthState.Unauthenticated) {
+                        if (mutableStateFlow.value !is RootNavState.Auth) {
+                            mutableStateFlow.update { RootNavState.Auth }
+                        }
+                    }
+                }
+        }
         combine(
             userDataRepository.authState,
             userDataRepository.settingsState,
--- a/core/network/src/commonMain/kotlin/org/mifos/mobile/core/network/utils/KtorInterceptor.kt
+++ b/core/network/src/commonMain/kotlin/org/mifos/mobile/core/network/utils/KtorInterceptor.kt
@@ -13,11 +13,14 @@ import io.ktor.client.HttpClient
 import io.ktor.client.plugins.HttpClientPlugin
 import io.ktor.client.request.HttpRequestPipeline
 import io.ktor.client.request.header
+import io.ktor.client.statement.HttpResponsePipeline
+import io.ktor.http.HttpStatusCode
 import io.ktor.util.AttributeKey
 import org.mifos.mobile.core.datastore.UserPreferencesRepository
 
 class KtorInterceptor(
     private val getToken: () -> String?,
+    private val onUnauthorized: (suspend () -> Unit)?,
 ) {
     companion object Plugin : HttpClientPlugin<Config, KtorInterceptor> {
         private const val HEADER_TENANT = "Fineract-Platform-TenantId"
@@ -39,17 +42,31 @@ class KtorInterceptor(
                     }
                 }
             }
+            scope.responsePipeline.intercept(HttpResponsePipeline.After) {
+                if (context.response.status == HttpStatusCode.Unauthorized) {
+                    runCatching {
+                        plugin.onUnauthorized?.invoke()
+                    }.onFailure { throwable ->
+                        throwable.printStackTrace()
+                    }
+                }
+                proceed()
+            }
         }
 
         override fun prepare(block: Config.() -> Unit): KtorInterceptor {
             val config = Config().apply(block)
-            return KtorInterceptor(config.getToken)
+            return KtorInterceptor(
+                getToken = config.getToken,
+                onUnauthorized = config.onUnauthorized,
+            )
         }
     }
 }
 
 class Config {
     lateinit var getToken: () -> String?
+    var onUnauthorized: (suspend () -> Unit)? = null
 }
 
 class KtorInterceptorRe(
--- a/core/network/src/commonMain/kotlin/org/mifos/mobile/core/network/di/NetworkModule.kt
+++ b/core/network/src/commonMain/kotlin/org/mifos/mobile/core/network/di/NetworkModule.kt
@@ -28,6 +28,10 @@ val NetworkModule = module {
             install(Auth)
             install(KtorInterceptor) {
                 getToken = { preferencesRepository.token.value }
+
+                onUnauthorized = suspend {
+                    preferencesRepository.logOut()
+                }
             }
         }
     }
--- a/feature/home/src/commonMain/kotlin/org/mifos/mobile/feature/home/HomeScreen.kt
+++ b/feature/home/src/commonMain/kotlin/org/mifos/mobile/feature/home/HomeScreen.kt
@@ -74,6 +74,10 @@ internal fun HomeScreen(
 ) {
     val state by viewModel.stateFlow.collectAsStateWithLifecycle()
 
+    LaunchedEffect(Unit) {
+        viewModel.handleAuthCheckOnResume()
+    }
+
     EventsEffect(viewModel.eventFlow) { event ->
         when (event) {
             is HomeEvent.Navigate -> {
--- a/feature/home/src/commonMain/kotlin/org/mifos/mobile/feature/home/HomeViewModel.kt
+++ b/feature/home/src/commonMain/kotlin/org/mifos/mobile/feature/home/HomeViewModel.kt
@@ -55,6 +55,8 @@ internal class HomeViewModel(
     ),
 ) {
 
+    private var isHandlingNetworkChange = false
+
     init {
         observeNetworkStatus()
     }
@@ -69,7 +71,9 @@ internal class HomeViewModel(
             networkMonitor.isOnline
                 .distinctUntilChanged()
                 .collect { isOnline ->
+                    isHandlingNetworkChange = true
                     sendAction(HomeAction.ObserveNetworkStatus(isOnline))
+                    isHandlingNetworkChange = false
                 }
         }
     }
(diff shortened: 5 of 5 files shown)


Nagarjuna0033:
@Kartikey15dem this looks good but let's take a scenario, suppose user session has expired and user on home screen and then navigate to other screens then this will work as it will perform network operation on that screen, so to handle this edge case, there is one api to check weather user is authorised or not by sending token to server. create a function for that and use `disposable effect` in home screen so whenever user enters into home screen that function will check weather user is authorized or not. If you have better approach, feel free to share.

Kartikey15dem:
> @Kartikey15dem this looks good but let's take a scenario, suppose user session has expired and user on home screen and then navigate to other screens then this will work as it will perform network operation on that screen, so to handle this edge case, there is one api to check weather user is authorised or not by sending token to server. create a function for that and use `disposable effect` in home screen so whenever user enters into home screen that function will check weather user is authorized or not. If you have better approach, feel free to share.

So, you don't want the user to logout if he is on the home screen and and the navigates to another screen and aren't there any api calls at the home screen itself. Can you explain me in detail for which cases this feature is required and for which cases this feature is not required like this one why is it so and what do you want after it will be checked that if the user is authorized or not at the home screen.

Kartikey15dem:
@Nagarjuna0033 
I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?

Nagarjuna0033:
> @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?

Yes you are right, proceed with that

Kartikey15dem:
> > @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?
> 
> Yes you are right, proceed with that

POST https://DomainName/api/v1/self/userdetails?access_token={access_token}- In this endpoint we will be passing the baseAuth token in the header,right? If yes , then I think passing the accessToken again as a query feels redundant and will cause unnecessary dependency for authrepository in HomeRespositoryImp to fetch the accessToken but if the can't be removed I think I will have to implement it this way only.

Nagarjuna0033:
> > > @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?
> > 
> > 
> > Yes you are right, proceed with that
> 
> POST [https://DomainName/api/v1/self/userdetails?access_token={access_token}-](https://DomainName/api/v1/self/userdetails?access_token=%7Baccess_token%7D-) In this endpoint we will be passing the baseAuth token in the header,right? If yes , then I think passing the accessToken again as a query feels redundant and will cause unnecessary dependency for authrepository in HomeRespositoryImp to fetch the accessToken but if the can't be removed I think I will have to implement it this way only.

@Kartikey15dem did you check the response without sending query? Let's see what they mention in the API docs.

Kartikey15dem:
> > > > @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?
> > > 
> > > 
> > > Yes you are right, proceed with that
> > 
> > POST [https://DomainName/api/v1/self/userdetails?access_token={access_token}-](https://DomainName/api/v1/self/userdetails?access_token=%7Baccess_token%7D-) In this endpoint we will be passing the baseAuth token in the header,right? If yes , then I think passing the accessToken again as a query feels redundant and will cause unnecessary dependency for authrepository in HomeRespositoryImp to fetch the accessToken but if the can't be removed I think I will have to implement it this way only.
> 
> @Kartikey15dem did you check the response without sending query?
@Nagarjuna0033 
Not yet as I thought it won't work but I will check once

Nagarjuna0033:
> > > > > @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?
> > > > 
> > > > 
> > > > Yes you are right, proceed with that
> > > 
> > > 
> > > POST [https://DomainName/api/v1/self/userdetails?access_token={access_token}-](https://DomainName/api/v1/self/userdetails?access_token=%7Baccess_token%7D-) In this endpoint we will be passing the baseAuth token in the header,right? If yes , then I think passing the accessToken again as a query feels redundant and will cause unnecessary dependency for authrepository in HomeRespositoryImp to fetch the accessToken but if the can't be removed I think I will have to implement it this way only.
> > 
> > 
> > @Kartikey15dem did you check the response without sending query?
> > @Nagarjuna0033
> > Not yet as I thought it won't work but I will check once

Yes, Api must require query as well as header. you can check it

Kartikey15dem:
> > > > > > @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?
> > > > > 
> > > > > 
> > > > > Yes you are right, proceed with that
> > > > 
> > > > 
> > > > POST [https://DomainName/api/v1/self/userdetails?access_token={access_token}-](https://DomainName/api/v1/self/userdetails?access_token=%7Baccess_token%7D-) In this endpoint we will be passing the baseAuth token in the header,right? If yes , then I think passing the accessToken again as a query feels redundant and will cause unnecessary dependency for authrepository in HomeRespositoryImp to fetch the accessToken but if the can't be removed I think I will have to implement it this way only.
> > > 
> > > 
> > > @Kartikey15dem did you check the response without sending query?
> > > @Nagarjuna0033
> > > Not yet as I thought it won't work but I will check once
> 
> Yes, Api must require query as well as header. you can check it

@Nagarjuna0033. 
So that means I have to proceed in this way only and no changes can be made from the backend for the redundant query.

Nagarjuna0033:
> > > > > > > @Nagarjuna0033 I think LaunchedEffect(Unit) is enough for our case as I think we should only check whenever we are navigating to the HomeScreen and then as long we remain on the screen we should not check otherwise there will be redundant api calls. What do you think?
> > > > > > 
> > > > > > 
> > > > > > Yes you are right, proceed with that
> > > > > 
> > > > > 
> > > > > POST [https://DomainName/api/v1/self/userdetails?access_token={access_token}-](https://DomainName/api/v1/self/userdetails?access_token=%7Baccess_token%7D-) In this endpoint we will be passing the baseAuth token in the header,right? If yes , then I think passing the accessToken again as a query feels redundant and will cause unnecessary dependency for authrepository in HomeRespositoryImp to fetch the accessToken but if the can't be removed I think I will have to implement it this way only.
> > > > 
> > > > 
> > > > @Kartikey15dem did you check the response without sending query?
> > > > @Nagarjuna0033
> > > > Not yet as I thought it won't work but I will check once
> > 
> > 
> > Yes, Api must require query as well as header. you can check it
> 
> @Nagarjuna0033. So that means I have to proceed in this way only and no changes can be made from the backend for the redundant query.

Yes no changes can be made from backend. send token and check in the postman if you have any doubt

Kartikey15dem:
I have made the requried changes. You can review it now.

Nagarjuna0033:
> I have made the requried changes. You can review it now.

Have you tested by changing password. change password and try to open in other device to check weather this is working or not.

Nagarjuna0033:
@kartikey15dem update the branch and upload the video

Kartikey15dem:
> @Kartikey15dem update the branch and upload the video

@Nagarjuna0033 
Updated the branch and added the video

Nagarjuna0033 (approved):

Nagarjuna0033 (approved):

cmp-navigation/src/commonMain/kotlin/cmp/navigation/rootnav/RootNavViewModel.kt
```diff
@@ -27,6 +28,18 @@ class RootNavViewModel(
 ) {
 
     init {
+        // Detect logout transitions
```
biplab1: Please remove these type of comments unless very important.

cmp-navigation/src/commonMain/kotlin/cmp/navigation/rootnav/RootNavViewModel.kt
```diff
@@ -27,6 +28,18 @@ class RootNavViewModel(
...
+        viewModelScope.launch {
+            userDataRepository.authState
+                .collect { authState ->
+                    // Navigate to Auth when user becomes unauthenticated
```
biplab1: Same here, please review. Also, remove such comments from elsewhere in this PR.

core/network/src/commonMain/kotlin/org/mifos/mobile/core/network/utils/KtorInterceptor.kt
```diff
@@ -39,17 +42,33 @@ class KtorInterceptor(
                     }
                 }
             }
+            scope.responsePipeline.intercept(HttpResponsePipeline.After) {
+                // Trigger onUnauthorized handler when response code is 401
```
biplab1: Same here, please review.

core/network/src/commonMain/kotlin/org/mifos/mobile/core/network/utils/KtorInterceptor.kt
```diff
@@ -39,17 +42,33 @@ class KtorInterceptor(
...
+                if (context.response.status == HttpStatusCode.Unauthorized) {
+                    runCatching {
+                        plugin.onUnauthorized?.invoke()
+                    }.onFailure { throwable ->
+                        // Handle the exception if needed
```
biplab1: Same here, please review.

biplab1 (approved):
Some minor changes requested. Rest looks good to me. This can be merged once the requested updates have been made.

Kartikey15dem:
@biplab1 
I have made the changes. Please approve it.

PR #3060, opened 2026-01-06
fix: language revert to system default after switching back from a custom language

amanna13:
### Before 
https://github.com/user-attachments/assets/3d04005e-2b66-4a29-81f6-9b82ada4f217

### After 
https://github.com/user-attachments/assets/3c5add3a-94a1-41ef-a161-5f0e01a2d909

- [x] Run the static analysis check `./gradlew check` or `ci-prepush.sh` to make sure you didn't break anything

----
### Explanation for the change 
**Problem** - The previous code didn’t trigger a proper configuration change to reset the language to the system default(null). `it?.let {...}` block is skipped entirely as a result UI stays in the previous locale. Even if this was fixed, the core issue was `MainActivity `extendng the `ComponentActivity`.  `ComponentActivity `**doesn't** :
- Override `attachBaseContext() `to apply the AppCompat locale wrapper
- Handle configuration changes in an AppCompat-aware way
- Handle Proper activity recreation and locale fallback mechanisms

Thus, it is migrated to `AppCompatActivity `which wraps the base context and applies Android's locale configuration rules. 

> [!NOTE] 
**Additional Fix - RTL Support:** : While the primary issue was language not reverting to the system default, using `AppCompatActivity` also enabled RTL support for languages like Urdu, which was previously not possible (Only changes strings- incomplete locale change)(_Check attachment_)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced locale handling to improve app language and regional settings management.

### DIFF
--- a/cmp-android/src/main/kotlin/cmp/android/app/MainActivity.kt
+++ b/cmp-android/src/main/kotlin/cmp/android/app/MainActivity.kt
@@ -10,8 +10,8 @@
 package cmp.android.app
 
 import android.os.Bundle
-import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.os.LocaleListCompat
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
@@ -33,9 +33,9 @@ import kotlin.getValue
  * This class is used to set the content view of the activity.
  *
  * @constructor Create empty Main activity
- * @see ComponentActivity
+ * @see AppCompatActivity
  */
-class MainActivity : ComponentActivity() {
+class MainActivity : AppCompatActivity() {
     /**
      * Called when the activity is starting.
      * This is where most initialization should go: calling [setContentView(int)] to inflate the activity's UI,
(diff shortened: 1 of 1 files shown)


cmp-android/src/main/kotlin/cmp/android/app/MainActivity.kt
```diff
@@ -33,9 +33,9 @@ import kotlin.getValue
  * This class is used to set the content view of the activity.
  *
  * @constructor Create empty Main activity
- * @see ComponentActivity
+ * @see AppCompatActivity
  */
-class MainActivity : ComponentActivity() {
+class MainActivity : AppCompatActivity() {
```
biplab1, line 38: @amanna13 Not fully convinced that switching to `AppCompatActivity` is the root fix here. Could we try solving the issue without changing the base activity and confirm whether that change is truly required?

biplab1 (changes requested):

cmp-android/src/main/kotlin/cmp/android/app/MainActivity.kt
amanna13: If we use `ComponentActivity` we need to implement the base Context manually, and also destroy the current activity and recreate it. Migrating to `AppCompatActivity` helps for AppCompat locale support by default and make the whole thing simpler. And already we are using `AppComaptDelegate` for locale changes.

biplab1: @amanna13 Let me look for an alternative solution without changing the base activity.

biplab1 (approved):
Looks good to me. This can be merged.

PR #3088, opened 2026-01-21
fix(cmp-shared):persist language changes for non-Android platforms

Kartikey15dem:
Fixes -[MM-537](https://mifosforge.jira.com/browse/MM-537)

## Description

On non-Android platforms (iOS and Desktop), changing the app language did not update the displayed text.
The selected language was not applied during onboarding nor when changed later via Settings, while Android worked as expected.

This PR fixes the issue by ensuring that the selected language is correctly applied at the platform level on iOS and Desktop, allowing Compose resources to recompose with the updated locale.

### Behavior Changes
| Before | After |
| :---: | :---: |
| Before | After |

| Before | After |
| :---: | :---: |
| Before | After |

---

## Root Cause

- Android correctly applies locale changes via platform APIs.
- Desktop did not update the JVM default locale when the language changed.
- iOS did not update the system language (`AppleLanguages`) when the app language was changed.
- Compose Multiplatform continued resolving strings using the previous locale on non-Android platforms.

---

### Desktop
- Applied the selected locale using `Locale.setDefault()` when the language changes.
- Connected the locale handler from `SharedApp` to the Desktop entry point.

### iOS
- Persisted the selected language using `NSUserDefaults` under the `AppleLanguages` key.
- Removed the override when the language is reset to default.

---

## Verification

- Language change verified during onboarding on Desktop and iOS.
- Language change verified via Settings on Desktop and iOS.
- UI updates correctly without app restart.
- No regression observed on Android.

---
[MM-537]: https://mifosforge.jira.com/browse/MM-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Desktop: changing language now takes effect immediately; empty selection reverts to system default.
  * iOS: selected language is persisted across launches; clearing the selection reverts to the device default.
  * Cross-platform: unified, active locale handling on both desktop and iOS for consistent language behavior.


biplab1:
Instead of gifs use videos since frames might be missed sometimes showing some critical transition behaviors.

niyajali:
@Kartikey15dem copy the exact implementation from the template project(https://github.com/openMF/kmp-project-template/blob/dev/cmp-desktop/src/jvmMain/kotlin/main.kt) and don't change any of the codebases from these module files

Kartikey15dem:
> @Kartikey15dem copy the exact implementation from the template project(https://github.com/openMF/kmp-project-template/blob/dev/cmp-desktop/src/jvmMain/kotlin/main.kt) and don't change any of the codebases from these module files

I didn't understood completely . Why this implementation was not there in mifos-mobile and what is the purpose of kmp-porject-template and which modules am I not supposed to change? Can you explain me a little bit more about my mistake so that I don't repeat it again. Also, is there any issue in looking bugs for iOS and desktop ?

niyajali:
@Kartikey15dem These files will be synced from the template. The syncing happens for these directories and modules. If changes do not match the template, the file will be updated during syncing.

Kartikey15dem:
> @Kartikey15dem These files will be synced from the template. The syncing happens for these directories and modules. If changes do not match the template, the file will be updated during syncing.

@niyajali 
Since the issue which is being solved in this PR is already being solved in kmp-project-template so I think it will be better for me to close this PR instead of copy-pasting the whole code if those changes will be merged later.

Kartikey15dem:
> > @Kartikey15dem These files will be synced from the template. The syncing happens for these directories and modules. If changes do not match the template, the file will be updated during syncing.
> 
> @niyajali 
> Since the issue which is being solved in this PR is already being solved in kmp-project-template so I think it will be better for me to close this PR instead of copy-pasting the whole code if those changes will be merged later.
@niyajali can you please confirm that should I close this PR as the implementations are already there in the kmp project template.

niyajali:
Why not just copy the exact implementation and update it here? Syncing could introduce many changes that would require significant rework.

Kartikey15dem:
> Why not just copy the exact implementation and update it here? Syncing could introduce many changes that would require significant rework.

Made the changes in main.kt for desktop

Kartikey15dem:
@niyajali 
If you no further changes are not required then this could me merged.

niyajali (approved):

PR #3101, opened 2026-02-01
feat: enhance error handling

Nagarjuna0033:
Fixes - [Jira-#527](https://mifosforge.jira.com/browse/MM-527)

| | |

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error messaging across authentication, accounts, and transactions with standardized, localized messages for network and server errors
  * Improved error feedback throughout the app for better user experience during account operations and transaction processing

### DIFF
--- a/core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/BeneficiaryRepositoryImp.kt
+++ b/core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/BeneficiaryRepositoryImp.kt
@@ -57,9 +57,9 @@ class BeneficiaryRepositoryImp(
                 val errorMessage = extractErrorMessage(e.response)
                 DataState.Error(Exception(errorMessage), null)
             } catch (e: IOException) {
-                DataState.Error(Exception("Network error: ${e.message ?: "Please check your connection"}"), null)
+                DataState.Error(Exception("Network error", e), null)
             } catch (e: ServerResponseException) {
-                DataState.Error(Exception("Server error: ${e.message}"), null)
+                DataState.Error(Exception("Server error", e), null)
             }
         }
     }
@@ -77,9 +77,9 @@ class BeneficiaryRepositoryImp(
                 val errorMessage = extractErrorMessage(e.response)
                 DataState.Error(Exception(errorMessage), null)
             } catch (e: IOException) {
-                DataState.Error(Exception("Network error: ${e.message ?: "Please check your connection"}"), null)
+                DataState.Error(Exception("Network error", e), null)
             } catch (e: ServerResponseException) {
-                DataState.Error(Exception("Server error: ${e.message}"), null)
+                DataState.Error(Exception("Server error", e), null)
             }
         }
     }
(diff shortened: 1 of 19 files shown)


core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/mapper/client/ClientMapper.kt
```diff
@@ -79,24 +79,24 @@ fun ClientClassificationResponseDto.toModel(): ClientClassification =
...
     )
 
 fun GenderResponseDto.toModel(): Gender =
     Gender(
         id = id,
         name = name,
-        active = active,
-        mandatory = mandatory,
+        active = active ?: false,
+        mandatory = mandatory ?: false
```
gururani-abhishek: One of the Gradle Checks is failing here. The Spotless/Ktlint is enforcing a trailing comma after the multi-line constructor parameters.

core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/BeneficiaryRepositoryImp.kt
mena-rizkalla, line 74: I know this is not a part from your ticket but i think using beneficiaryId!! will make the app crash if the id is null, if this id is required we shouldn't make it nullable or if it must we can handle the null case explicitly with a DataState.Error

feature/auth/src/commonMain/kotlin/org/mifos/mobile/feature/auth/login/LoginViewModel.kt
```diff
@@ -88,45 +92,56 @@ class LoginViewModel(
...
-                        showOverlay = false,
-                        dialogState = LoginState.DialogState.Error(action.loginResult.message),
-                        userNameError = Res.string.feature_sign_in_username_error,
-                        passwordError = Res.string.feature_sign_in_password_error,
-                    )
+        viewModelScope.launch {
+            when (action.loginResult) {
+                is DataState.Error -> {
+                    val errorMsg =
+                        if (action.loginResult.exception.cause is ServerResponseException) {
```
mena-rizkalla, line 99: why you use "ServerResponseException" i can see this is from ktor library why viewmodel should know which network library we use ? maybe we need to change it in the future

feature/auth/src/commonMain/kotlin/org/mifos/mobile/feature/auth/login/LoginViewModel.kt
```diff
@@ -88,45 +92,56 @@ class LoginViewModel(
...
-                    officeName = user.officeName.orEmpty(),
-                    password = state.password,
-                )
-                viewModelScope.launch {
-                    userPreferencesRepositoryImpl.updateUser(userData)
-                    userPreferencesRepositoryImpl.setIsAuthenticated(true)
+                is DataState.Success -> {
+                    updateState { it.copy(showOverlay = false) }
+                    val user = action.loginResult.data
+                    val userData = UserData(
```
mena-rizkalla, line 126: i know this is not a part of your pr but we can make a new ticket for it as i think this part should belong to the mapper you made before

mena-rizkalla (changes requested):

core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/BeneficiaryRepositoryImp.kt
Nagarjuna0033: @mena-rizkalla The id will never be nulll.

itsPronay (approved):

mena-rizkalla (commented):
LGTM, but i think it's better not to use "ServerResponseException" in the viewmodel as it's from ktor library and it's better if  viewmodel doesn't depend on the data layer

Issue #577, opened 2018-02-06
feat:Reporting of crash while using app absent.


ivary43:
**Summary:**
Feature of reporting an app crash is currently absent in app.

**Expected behaviour:**

We should send the reports to our organisation desk or somewhere more suitable So that we could analyze and fix the bugs for future updates after the release of the app.

**Observed behaviour:**

This feature is currently not present in the app.

ivary43:
@therajanmaurya, @dilpreet96  can I implement this feature?

miPlodder:
@ivary43 This feature is already implemented in the App. Although the API is not added (#820 )

luckyman20:
@miPlodder Thanks for reporting.

PR #3134, opened 2026-04-03
refactor(core): introduce domain exceptions for ui error handling

mena-rizkalla:
- Introduce `MifosException` sealed class to categorize errors into `ServerError`, `ClientError`, `NetworkError`, and `GenericError`.
- Add `ExceptionMapper` utility to convert `Throwable` and Ktor exceptions into `MifosException` types, including a suspend version for async error extraction.
- Update `asDataStateFlow` extension to support custom exception mapping.
- Refactor all repository implementations (`BeneficiaryRepositoryImp`, `LoanRepositoryImp`, `SavingsAccountRepositoryImp`, etc.) to use the new exception mapping logic.
- Update ViewModels across all features to handle specific `MifosException` types instead of raw IO or Ktor exceptions for UI state management.

Fixes - [Jira-#577](https://mifosforge.jira.com/browse/MM-577)

## Summary by CodeRabbit

* **Refactor**
  * Centralized error handling across repositories and view models for consistent network vs server error classification
  * User-facing messages now distinguish network, server, client, and generic errors more reliably
  * Reduced silent/ignored failures so error states propagate to UI collectors, improving refresh/error indicators and status dialogs

### DIFF
--- a/core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/AccountsRepositoryImp.kt
+++ b/core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/AccountsRepositoryImp.kt
@@ -11,24 +11,21 @@ package org.mifos.mobile.core.data.repositoryImpl
 
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.map
 import org.mifos.mobile.core.common.DataState
-import org.mifos.mobile.core.common.asDataStateFlow
 import org.mifos.mobile.core.data.mapper.accounts.toModel
 import org.mifos.mobile.core.data.repository.AccountsRepository
-import org.mifos.mobile.core.data.util.toMifosException
 import org.mifos.mobile.core.model.entity.client.ClientAccounts
 import org.mifos.mobile.core.network.DataManager
 
 class AccountsRepositoryImp(
     private val dataManager: DataManager,
-    private val ioDispatcher: CoroutineDispatcher,
-) : AccountsRepository {
+    ioDispatcher: CoroutineDispatcher,
+) : BaseRepository(ioDispatcher), AccountsRepository {
 
     override fun loadAccounts(clientId: Long?, accountType: String?): Flow<DataState<ClientAccounts>> {
         return dataManager.clientsApi.getAccounts(clientId!!, accountType)
             .map { it.toModel() }
-            .asDataStateFlow(Throwable::toMifosException).flowOn(ioDispatcher)
+            .asDataState()
     }
 }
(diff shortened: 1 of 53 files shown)


core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/AccountsRepositoryImp.kt
```diff
@@ -28,6 +29,6 @@ class AccountsRepositoryImp(
     override fun loadAccounts(clientId: Long?, accountType: String?): Flow<DataState<ClientAccounts>> {
         return dataManager.clientsApi.getAccounts(clientId!!, accountType)
             .map { it.toModel() }
-            .asDataStateFlow().flowOn(ioDispatcher)
+            .asDataStateFlow(Throwable::toMifosException).flowOn(ioDispatcher)
     }
```
niyajali, line 30: We can map this directly in the `toDataStateFlow` extension function, so we won't need to update each repository method individually, alternatively, we could create an overload to keep it backward compatible.
We also need to make sure we're catching server-side errors properly and mapping them to the right classes. To do that, we'll need a full list of possible error responses from the backend, either check the web app or ask Victor if there's any documentation or reference for those.

In the meantime, check the `NetworkResult` class in the `core-base/network` module. it should give a good starting point for how errors are currently being handled and what patterns we can follow.

niyajali (changes requested):

core/data/src/commonMain/kotlin/org/mifos/mobile/core/data/repositoryImpl/AccountsRepositoryImp.kt
biplab1: @mena-rizkalla You could take a look at web-app for Http Status Error codes at: https://github.com/openMF/web-app/blob/dev/src/app/core/http/error-handler.interceptor.ts, Lines: 118-169

| Status Code | Error Type | Message Source | Handling Behavior |
|-------------|------------|----------------|-------------------|
| 400 | Bad Request | `response.error.errors[0].defaultUserMessage` or `response.error.errors[0].developerMessage` | Shows alert with `"Invalid parameters were passed in the request!"` fallback |
| 401 | Authentication Error | Fixed message | Shows `"Invalid User Details. Please try again!"` |
| 403 | Unauthorized | `errorMessage` or fixed message | Shows `"You are not authorized for this request!"` (special handling for invalid one-time token) |
| 404 | Not Found | `errorMessage` or fixed message | Shows `"Resource does not exist!"` (silently handles client image 404s) |
| 500 | Internal Server Error | Fixed message | Shows `"Internal Server Error. Please try again later."` |
| 501 | Not Implemented | Translation keys | Shows translated `"Not Implemented Error"` message |

Also API uses standard Http Status Codes, https://fineract.apache.org/docs/legacy/index.html

| Status Code | Meaning | Typical Use Case |
|-------------|---------|------------------|
| 200 | OK | Success |
| 400 | Bad Request | Validation errors, invalid parameters |
| 401 | Authentication Error | Invalid credentials |
| 403 | Unauthorized | Insufficient permissions |
| 404 | Not Found | Resource doesn't exist |
| 409 | Conflict | Concurrent modification, locked resources |
| 500 | Internal Server Error | Unexpected server errors |

image

